PR #6, opened 2024-05-17
chore: improve dependency handling and remove the mongodb driver

kmruiz:
### Checklist
- [x] Documentation is changed or added.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,6 +38,11 @@ subprojects {
         val testImplementation by configurations
         val compileOnly by configurations
 
+        configurations.named("runtimeClasspath").configure {
+            exclude("org.jetbrains.kotlin")
+            exclude("org.jetbrains.kotlinx")
+        }
+
         compileOnly(rootProject.libs.kotlin.stdlib)
         testImplementation(rootProject.libs.testing.jupiter.engine)
         testImplementation(rootProject.libs.testing.jupiter.vintage.engine)
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,10 +14,8 @@ diktat-plugin="1.0.1"
 jmhreport-plugin="0.9.0"
 
 # Library dependencies
-bson="5.0.0"
-mongodb-driver="4.11.0"
 kotlin-stdlib="1.9.22"
-kotlinx-coroutines="1.8.0-RC2"
+kotlinx-coroutines="1.7.3"
 jupiter="5.10.2"
 mockito="5.11.0"
 mockito-kotlin="5.3.1"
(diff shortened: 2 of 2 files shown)


packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -0,0 +1,129 @@
...
+    override suspend fun <T : Any> findAll(
+        namespace: Namespace,
+        query: Document,
+        result: KClass<T>,
+        limit: Int,
+        timeout: Duration
+    ) = withContext(Dispatchers.IO) {
+        runQuery(
+            """db.getSiblingDB("${namespace.database}")
+                 .getCollection("${namespace.collection}")
```
addaleax, line 79: Do you want to account for escaping here? (This might already be addressed by your updates locally, but just checking)

Otherwise some kind of TODO might be nice, even if this is still WIP :)

packages/mongodb-access-adapter/datagrip-access-adapter/src/test/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/IntegrationTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    ParameterResolver {
+    private val namespace = ExtensionContext.Namespace.create(IntegrationTestExtension::class.java)
+    private val containerKey = "CONTAINER"
+    private val projectKey = "PROJECT"
+    private val driverKey = "DRIVER"
+    private val versionKey = "VERSION"
+
+    override fun beforeAll(context: ExtensionContext?) {
+        val annotation = context!!.requiredTestClass.getAnnotation(IntegrationTest::class.java)
+        val container = MongoDBContainer("mongo:${annotation.mongodb.versionString}-jammy")
```
addaleax, line 69: Just a clarifying question: container usage = We're not planning to run tests on platforms other than Linux?

kmruiz: I'm not sure I follow, they should run on any environment with Docker/Podman (I'm running them on my Mac with Podman right now). Do you see any concern here?

gradle/diktat.yml
```diff
@@ -17,4 +17,8 @@
 - name: TOO_MANY_PARAMETERS
   enabled: true
   configuration:
-    maxParameterListSize: '10'
\ No newline at end of file
+    maxParameterListSize: '10'
+- name: FILE_NAME_INCORRECT
+  enabled: false
```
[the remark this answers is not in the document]
kmruiz: Added a comment for each rule explaining why it's disabled / enabled.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: I added some escaping now, thanks for pointing out! I was escaping only when converting a string to a namespace using the `.toNs` method, but not in the namespace constructor. Now I fixed it and it should be safe.

PR #7, opened 2024-06-11
chore: flush telemetry when IDE closes INTELLIJ-11

kmruiz:
## Description

This PR makes the TelemetryService subscribe to an event that IntelliJ emits when it's going to close. Before the IDE closes, we flush any enqueued telemetry events to Segment.

I also cleaned up a little bit how the dependency injection works, by using Application Services whenever it makes sense, and simplifying setting up mocks in tests.

### Checklist


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ MongoDB plugin for IntelliJ IDEA.
 ## [Unreleased]
 
 ### Added
+* [INTELLIJ-11](https://jira.mongodb.org/browse/INTELLIJ-11): Flush pending analytics events before closing the IDE.
```
addaleax, line 8: Is this the general format we'll be using for changelog management? I'm a bit worried that this will lead to merge conflicts once there's more than one person working on this product 🙂

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
```diff
@@ -15,7 +16,7 @@ import kotlinx.coroutines.launch
 class ActivatePluginPostStartupActivity(private val cs: CoroutineScope) : StartupActivity, DumbAware {
     override fun runActivity(project: Project) {
         cs.launch {
-            val pluginActivated = project.getService(PluginActivatedProbe::class.java)
+            val pluginActivated = ApplicationManager.getApplication().getService(PluginActivatedProbe::class.java)
```
addaleax, line 19: Is there some documentation where I could learn about the difference between the two/why `ApplicationManager` is preferred for accessing services? Naïvely I would assume that `ApplicationManager.getApplication()` returns a singleton, which admittedly doesn't make a ton of sense to me in combination with dependency/service injection mechanisms

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/IntegrationTestExtensions.kt
```diff
@@ -0,0 +1,153 @@
...
+ * @param arch
+ * @param jvmVendor
+ * @param jvmVersion
+ * @param buildVersion
+ * @param applicationName
+ * @return A new mocked RuntimeInformationService
+ */
+internal fun mockRuntimeInformationService(
+    userId: String = "123456",
+    osName: String = "Winux OSX",
```
addaleax, line 117: 😬

kmruiz: I knew you were going to hate it and I'm proud of it.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
kmruiz: There is no much documentation, to be fair, it's only here: https://plugins.jetbrains.com/docs/intellij/plugin-services.html

But basically, you are right in that ApplicationManager.getApplication is a singleton (potentially). The main reason to use the injection through an application or a project it is that if the service is a light service, the actual implementation of the service can change in runtime. An actual real example is when upgrading the plugin, that can be done without restarting the IDE.

The difference between an Application service and a Project service is the scope of:

* The coroutines
* The service itself

In an Application service, the scope of the coroutine is shared across projects. If you have 3 java projects open, and you use the Application coroutine scope, the underlying thread pool can be shared across all projects, reducing the number of threads needed.

In a Project service, you will have a coroutine scope for each project and also a service instance for each project. In some cases this might be desirable (for example, when you are parsing code), but for more generic use cases like telemetry, connecting to MongoDB, and so on is better to just have one single instance.

CHANGELOG.md
kmruiz: Merge conflicts in the CHANGELOG file you mean? Yeah I mean, that's an option, but I don't think that should be really troublesome. It should be a trivial merge conflict to fix.

This is the recommended way (from JetBrains at least) to keep track of the plugin changelog, as they already have a plugin that handles everything integrated (https://github.com/JetBrains/gradle-changelog-plugin).

addaleax: Alright – I think we've very intentionally tried to avoid this for Compass and mongosh, but if we're being diligent here and are okay with having to resolve merge conflicts, this is all fine by me 🙂

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
addaleax: (just going to leave this unresolved because I think it's helpful)

addaleax (approved):

PR #8, opened 2024-06-12
feat(telemetry): allow to disable telemetry through settings INTELLIJ-12

kmruiz:
## Description

Notifies a plugin user, the first time the plugin is activated, that we use anonymous telemetry. In the notification balloon there is an action to disable the telemetry from there. In the plugin settings there is a checkbox that allows us to toggle telemetry any time.

signal-2024-06-12-14-43-50-481
signal-2024-06-12-14-43-50-481-1

### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
```diff
@@ -1,13 +1,34 @@
+/**
+ * These class implement the balloon that shows the first time that the plugin is activated.
```
addaleax, line 2: typo?

```suggestion
 * These classes implement the balloon that shows the first time that the plugin is activated.
```

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/settings/PluginSettings.kt
```diff
@@ -0,0 +1,46 @@
+/**
+ * Settings state for the plugin. These classes are responsible for the persistence of the plugin
+ * settings.
```
addaleax, line 3: How "broad" is the scope of settings in IntelliJ? Is every setting defined here automatically visible and configurable to the user, or is it more like Compass, where it's fine for us to use some settings as easy ways of persisting small bits of data?

packages/jetbrains-plugin/src/main/resources/messages/TelemetryBundle.properties
```diff
@@ -0,0 +1,6 @@
+notification.group.name=MongoDB telemetry
+notification.title=MongoDB plugin telemetry
+notification.message=Anonymous telemetry is enabled by default because it helps us improve the plugin.
+action.disable-telemetry=Disable Telemetry
+settings.telemetry-collection-tooltip=Allow the collection of anonymous diagnostics and usage telemetry data to help improve the product.
+settings.telemetry-collection-checkbox=Enable telemetry
```
addaleax, line 6: Do we need to/can we include links to https://www.mongodb.com/legal/privacy/privacy-policy?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/settings/PluginSettings.kt
kmruiz: It's more like Compass. So you have the "settings state" that is persisted and it's not related anyhow to the settings view. How you want to handle these settings is up to you. There is a class (com.mongodb.jbplugin.settings.PluginSettingsConfigurable) that specifies how this settings can be configured in the Settings UI and it's not mandatory to add all of them, only whichever are relevant. 

To give you an example, `hasTelemetryOptOutputNotificationBeenShown` is not set up in the Settings page, but is set up in the initial activity (com.mongodb.jbplugin.ActivatePluginPostStartupActivity).

packages/jetbrains-plugin/src/main/resources/messages/TelemetryBundle.properties
kmruiz: That's a good point. I'm going to add a link on the notification and on the settings page to the privacy policy.

addaleax (approved):

PR #9, opened 2024-06-17
chore: send telemetry when connecting to a cluster INTELLIJ-13

kmruiz:
### Checklist


packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -33,60 +37,90 @@ import kotlinx.coroutines.withTimeout
...
+    override suspend fun countAll(
+        namespace: Namespace,
+        query: Bson,
+        timeout: Duration,
+    ) = withContext(Dispatchers.IO) {
+        runQuery(
+            """
+            db.getSiblingDB("${namespace.database}")
+                 .getCollection("${namespace.collection}")
+                 .countDocuments(${query.toJson()})
```
addaleax, line 113: Do we want a TODO ticket for adding proper escaping everywhere?

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
```diff
@@ -14,19 +16,95 @@ import org.bson.Document
...
+                    ".([01]?[0-9][0-9]?|2[0-4][0-9]|25[0-5])" +
+                    "|0.0.0.0" +
+                    "|(?:0*:)*?:?0*1" +
+                    ")$",
+            )
+        private val digitalOceanRegex = Regex(""".*\.mongo\.ondigitalocean\.com$""")
+        private val cosmosDbRegex = Regex(""".*\.cosmos\.azure\.com$""")
+        private val docDbRegex = Regex(""".*docdb(-elastic)?\.amazonaws\.com$""")
+
+        override suspend fun queryUsingDriver(from: MongoDbDriver): BuildInfo {
```
addaleax, line 65: `MongoDbDriver` can still be the mongosh-based DataGrip interface, right, not just the Java driver?

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -33,60 +37,90 @@ import kotlinx.coroutines.withTimeout
...
+    private val jsonWriterSettings = JsonWriterSettings.builder()
+.outputMode(JsonMode.EXTENDED)
+.indent(false)
+.build()
+
+    private fun Bson.toJson(): String = this.toBsonDocument().toJson(
+            jsonWriterSettings,
+        )
+
+    override suspend fun serverUri(): URI = URI.create(dataSource.url!!)
```
addaleax, line 52: This `URI` won't have a `host` property if it doesn't parse as a valid URI (e.g. multiple hosts). Should we really be using `URI` here?

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -33,60 +37,90 @@ import kotlinx.coroutines.withTimeout
...
-            timeout
-        )[0]
-    }
+        timeout: Duration,
+    ): T =
+        withContext(
+            Dispatchers.IO,
+        ) {
+            runQuery(
+                """db.runCommand(${command.toJson()})""",
```
addaleax, line 63: We're using this for `buildInfo`, which I'm pretty sure always needs to run against `admin`. Should we explicitly pass the db here somewhere? It can still default to `admin`, but that should probably still be an argument to the Kotlin `runCommand()` wrapper

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/LogMessage.kt
```diff
@@ -22,11 +22,20 @@ import com.intellij.openapi.components.Service
...
-    fun put(key: String, value: Any): LogMessageBuilder {
+    fun put(
+        key: String,
+        value: Any,
+    ): LogMessageBuilder {
         properties[key] = value
         return this
     }
 
+    inline fun <reified T : TelemetryEvent> merge(event: T): LogMessageBuilder {
```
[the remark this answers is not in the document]
kmruiz: Yeah, that's a great point. I think mergeTelemetryEventProperties is the best option. Fixing this!

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: Oh no, I just forgot about it (shame on me). Fixing it.

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
kmruiz: I'm not sure I understand the question. MongoDbDriver is a facade over a driver (like a DataService) and in some tests we use the real MongoDB driver and in production we use the DataGrip based driver.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: That's a good point. Maybe we can use the ConnectionString class from the Java driver.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: True true, I'm adding the database to the runCommand wrapper.

packages/mongodb-access-adapter/src/test/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfoTest.kt
```diff
@@ -1,22 +1,48 @@
...
+                    false,
+                    false,
+                    false,
+                    false,
+                    false,
+                    false,
+                    false,
+                    false,
+                    null,
+                    "mongodb://localhost",
```
addaleax, line 39: I think it's worth adding tests for things like multiple hostnames, hostnames that include ports, IPv6, etc. here :)

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
```diff
@@ -14,19 +16,96 @@ import org.bson.Document
...
+        private val atlasRegex = Regex(""".*\.mongodb(-dev|-qa|-stage)?\.net$""")
+        private val atlasStreamRegex = Regex("""^atlas-stream-.+""")
+        private val isLocalhostRegex =
+            Regex(
+                "^(localhost" +
+                    "|127.([01]?[0-9][0-9]?|2[0-4][0-9]|25[0-5])" +
+                    ".([01]?[0-9][0-9]?|2[0-4][0-9]|25[0-5])" +
+                    ".([01]?[0-9][0-9]?|2[0-4][0-9]|25[0-5])" +
+                    "|0.0.0.0" +
+                    "|(?:0*:)*?:?0*1" +
```
addaleax, line 58: The IPv6 check here needs to account for [brackets] around the IP, right?

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -33,60 +36,92 @@ import kotlinx.coroutines.withTimeout
...
+    override suspend fun countAll(
+        namespace: Namespace,
+        query: Bson,
+        timeout: Duration,
+    ) = withContext(Dispatchers.IO) {
+        runQuery(
+            """
+            db.getSiblingDB("${namespace.database}")
+                 .getCollection("${namespace.collection}")
+                 .countDocuments(EJSON.parse(`${query.toJson()}`))
```
addaleax, line 114: This is definitely beeter, but most queries here still have escaping issues (e.g.: the query could contain a literal `` ` ``, a database or collection name could include newlines, etc.)

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
kmruiz: It seems that the regex works differently in JS and Kotlin here. The [::1] IP works in mongodb-build-info, but had to fix it here 🤷

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
addaleax: > The query is a JSON document which is escaped when using toJson, so it should be safe.

So `` `${query.toJson()}` `` is not actuall backticks surrounding the raw JSON String? Can you point me to where that escaping is happening?

> com.mongodb.jbplugin.accessadapter.Namespace.Companion#invoke

Okay, yeah, that makes sense, but ... it's honestly pretty far from obvious how this works. Would it be possible to name this so that `namespace.database` gives you the database name as a Java String and `namespace.escapedDatabase` gives you the escaped version? That is a bit more verbose but won't give every reader of this code a reason to scratch their heads 🙂

kmruiz: That makes sense, I'm wondering how to do it in a way that is obvious which version to use, so we don't use the unescaped version in code.

kmruiz: I changed the name of the properties to `escaped` and added the non escaped versions. To avoid misuse of the non escaped versions, I marked the API as delicate and needs to be opted in when used using an annotation so it won't compile.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -33,60 +37,96 @@ import kotlinx.coroutines.withTimeout
...
-            timeout
-        ).getOrNull(0)
-    }
+        timeout: Duration,
+    ): T? =
+        withContext(Dispatchers.IO) {
+            runQuery(
+                """db.getSiblingDB("${namespace.database.encodeForJs()}")
+                 .getCollection("${namespace.collection.encodeForJs()}")
+                 .findOne(EJSON.parse(`${query.toJson()}`), EJSON.parse(`${options.toJson()}`))
```
addaleax, line 86: Does this work if the query contains a `` ` ``? Should this be 
```suggestion
                 .findOne(EJSON.parse("${query.toJson().encodeForJs()}"), EJSON.parse("${options.toJson()}.encodeForJs()")) 
```

kmruiz: Hmmm toJson already encodes the BSON document using the JsonWriterSettings, so that part should be safe. About using double quotes or back quotes, I think both should be fine? But I can change it to double quotes.

addaleax: I don't care about the specific quotes, just making sure that we can't end up with e.g. `` EJSON.parse(`{ "foo": "bar`baz" }`) `` here (because JSON writers in general don't need to escape backticks in strings)

addaleax (approved):

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: Makes sense, I'll change it to double quotes to make sure we don't get some weird scenario where someone can inject a query here.

Thanks!

addaleax: Ok, but now you definitely have issues because the JSON text will invariably contain double quotes (hence the suggestion to escape those, so we get reliably and "obviously correct" injections)

kmruiz: Yeah, you are right, I didn't realise. I'm going to force encoding the JSON that we convert using the OWASP encoder and use double quotes. This way it should be entirely safe.

PR #11, opened 2024-06-27
chore: add a placeholder toolbar if we detect code refering to the Java Driver in the editor INTELLIJ-15

kmruiz:
### Checklist


.run/Start IDE with Plugin.run.xml
```diff
@@ -0,0 +1,24 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Start IDE with Plugin" type="GradleRunConfiguration" factoryName="Gradle" folderName="Plugin">
+    <ExternalSystemSettings>
+      <option name="executionName" />
+      <option name="externalProjectPath" value="$PROJECT_DIR$/packages/jetbrains-plugin" />
```
addaleax, line 5: Does Jetbrains genuinely use `$foobar$` as a substitution syntax? 😨

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/idea/IdeaFrame.kt
```diff
@@ -0,0 +1,93 @@
...
+        this.closeAllFiles()
+
+        runJs(
+            """
+            importPackage(com.intellij.openapi.fileEditor)
+            importPackage(com.intellij.openapi.vfs)
+            importPackage(com.intellij.openapi.wm.impl)
+            importClass(com.intellij.openapi.application.ApplicationManager)
+            
+            const path = '$path'
```
addaleax, line 37: Should this be escaped?

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/idea/IdeaFrame.kt
```diff
@@ -0,0 +1,93 @@
...
+ * @param remoteRobot
+ * @param remoteComponent
+ */
+@FixtureName("Idea frame")
+@DefaultXpath("IdeFrameImpl type", "//div[@class='IdeFrameImpl']")
+class IdeaFrame(
+    remoteRobot: RemoteRobot,
+    remoteComponent: RemoteComponent,
+) : CommonContainerFixture(remoteRobot, remoteComponent) {
+    fun openFile(path: String) {
```
addaleax, line 27: I'm not sure where this is being used, do we need to worry about e.g. path traversal issues?

kmruiz: It's only for tests, it's in the test folder and likely it's going to be always hardcoded. The worst thing that can happen is a failing test. It's not bundled anyhow with the plugin.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/idea/IdeaFrame.kt
kmruiz: It's only for testing so I don't think it's that bad, but yeah I can encode it just in case we don't break anything during a test.

.run/Start IDE with Plugin.run.xml
kmruiz: I didn't realise TBF, but yeah it seems they use it for templates.

PR #21, opened 2024-08-01
chore: add atlas host to new connection events

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
@@ -19,6 +19,7 @@ internal enum class TelemetryProperty(
     val publicName: String,
 ) {
     IS_ATLAS("is_atlas"),
+    ATLAS_HOST("atlas_host"),
     IS_LOCAL_ATLAS("is_local_atlas"),
     IS_LOCALHOST("is_localhost"),
     IS_ENTERPRISE("is_enterprise"),
@@ -72,6 +73,7 @@ internal sealed class TelemetryEvent(
      * @param isLocalhost
      * @param isEnterprise
      * @param isGenuine
+     * @param atlasHost
      * @param nonGenuineServerName
      * @param serverOsFamily
      * @param version
@@ -83,6 +85,7 @@ internal sealed class TelemetryEvent(
         isLocalhost: Boolean,
         isEnterprise: Boolean,
         isGenuine: Boolean,
+        atlasHost: String?,
         nonGenuineServerName: String?,
         serverOsFamily: String?,
         version: String?,
@@ -95,6 +98,7 @@ internal sealed class TelemetryEvent(
                     TelemetryProperty.IS_LOCALHOST to isLocalhost,
                     TelemetryProperty.IS_ENTERPRISE to isEnterprise,
                     TelemetryProperty.IS_GENUINE to isGenuine,
+                    TelemetryProperty.ATLAS_HOST to (atlasHost ?: ""),
                     TelemetryProperty.NON_GENUINE_SERVER_NAME to (nonGenuineServerName ?: ""),
                     TelemetryProperty.SERVER_OS_FAMILY to (serverOsFamily ?: ""),
                     TelemetryProperty.VERSION to (version ?: ""),
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
@@ -19,7 +19,7 @@ internal enum class TelemetryProperty(
     val publicName: String,
 ) {
     IS_ATLAS("is_atlas"),
-    ATLAS_HOST("atlas_host"),
+    ATLAS_HOST("atlas_hostname"),
     IS_LOCAL_ATLAS("is_local_atlas"),
     IS_LOCALHOST("is_localhost"),
     IS_ENTERPRISE("is_enterprise"),
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
@@ -92,18 +92,24 @@ internal sealed class TelemetryEvent(
     ) : TelemetryEvent(
             name = "new-connection",
             properties =
-                mapOf(
-                    TelemetryProperty.IS_ATLAS to isAtlas,
-                    TelemetryProperty.IS_LOCAL_ATLAS to isLocalAtlas,
-                    TelemetryProperty.IS_LOCALHOST to isLocalhost,
-                    TelemetryProperty.IS_ENTERPRISE to isEnterprise,
-                    TelemetryProperty.IS_GENUINE to isGenuine,
-                    TelemetryProperty.ATLAS_HOST to (atlasHost ?: ""),
-                    TelemetryProperty.NON_GENUINE_SERVER_NAME to (nonGenuineServerName ?: ""),
-                    TelemetryProperty.SERVER_OS_FAMILY to (serverOsFamily ?: ""),
-                    TelemetryProperty.VERSION to (version ?: ""),
-                ),
-        )
+            mapOf(
+                TelemetryProperty.IS_ATLAS to isAtlas,
+                TelemetryProperty.IS_LOCAL_ATLAS to isLocalAtlas,
+                TelemetryProperty.IS_LOCALHOST to isLocalhost,
+                TelemetryProperty.IS_ENTERPRISE to isEnterprise,
+                TelemetryProperty.IS_GENUINE to isGenuine,
+                TelemetryProperty.NON_GENUINE_SERVER_NAME to (nonGenuineServerName ?: ""),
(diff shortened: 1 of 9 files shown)


packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
```diff
@@ -138,7 +138,7 @@ data class BuildInfo(
                 false,
                 false,
                 null,
-                "",
+                ConnectionString("mongodb://localhost"),
```
himanshusinghs, line 141: Just confirming, does it mean that by default the build info will be reported for `mongodb://localhost`. It probably does get overridden when connected, I am still wondering if making this nullable would be better?

himanshusinghs (approved):
Looking good, I left a small question and learnt about takeIf <3

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/slice/BuildInfo.kt
kmruiz: You are right. nullable would be a good option here for disconnected use cases. Let me change it.

Thanks a lot!

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/DataGripBasedReadModelProvider.kt
```diff
@@ -15,9 +15,13 @@ import com.mongodb.jbplugin.accessadapter.MongoDbDriver
...
 import com.mongodb.jbplugin.accessadapter.Slice
 import com.mongodb.jbplugin.accessadapter.datagrip.adapter.DataGripMongoDbDriver
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+
 import kotlinx.coroutines.runBlocking
 
-private typealias MapOfCachedValues = MutableMap<String, CachedValue<*>>
+private typealias MapOfCachedValues = ConcurrentMap<String, CachedValue<*>>
```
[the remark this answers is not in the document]
kmruiz: This map is used for all caches, both for parsed queries and new connections. It happened when I tried to open a new connection in debug mode while in another thread we were parsing a query. It shouldn't happen _a lot_, probably on slow computers in might be more common, but using a ConcurrentMap is the safest option and it doesn't change anyhow the API of the ReadModelProvider.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/TelemetryEvent.kt
```diff
@@ -95,6 +98,7 @@ internal sealed class TelemetryEvent(
                     TelemetryProperty.IS_LOCALHOST to isLocalhost,
                     TelemetryProperty.IS_ENTERPRISE to isEnterprise,
                     TelemetryProperty.IS_GENUINE to isGenuine,
+                    TelemetryProperty.ATLAS_HOST to (atlasHost ?: ""),
```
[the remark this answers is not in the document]
kmruiz: I think it's fine to send a null value too. The idea was to make sure that all events of the same type send all properties, so it's easier to reason about them. But I think it's fair to not send this one empty, so I'll change it.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbe.kt
```diff
@@ -58,6 +57,7 @@ class NewConnectionActivatedProbe : DatabaseSessionStateListener {
                 isGenuine = serverInfo.isGenuineMongoDb,
                 nonGenuineServerName = serverInfo.nonGenuineVariant,
                 serverOsFamily = serverInfo.buildEnvironment["target_os"],
+                atlasHost = serverInfo.serverUrl!!.hosts[0].takeIf { serverInfo.isAtlas && !serverInfo.isLocalAtlas },
```
alenakhineika, line 60: - What `serverInfo.isAtlas` checks? is it `atlasVersion` || `checking the urls matches atlas url` as we do for other tools? We also don't look for serverInfo.isLocalAtlas.
- Does atlasHost has port in it? I think normally hosts[0] has both `host:port`. We need to send only host in telemetry
- Is serverUrl has resolved URL?

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/TelemetryServiceTest.kt
```diff
@@ -53,6 +53,7 @@ internal class TelemetryServiceTest {
                 nonGenuineServerName = null,
                 serverOsFamily = null,
                 isLocalAtlas = false,
+                atlasHost = null,
```
alenakhineika, line 56: Here you have `isAtlas = true` so atlasHost should be a string hostname.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
```diff
@@ -68,7 +68,8 @@ internal abstract class NewConnectionActivatedProbeTest(
                     event.properties[TelemetryProperty.IS_LOCALHOST] == isLocalhost &&
                     event.properties[TelemetryProperty.IS_ENTERPRISE] == isEnterprise &&
                     event.properties[TelemetryProperty.IS_GENUINE] == isGenuine &&
-                    event.properties[TelemetryProperty.VERSION] == version
+                    event.properties[TelemetryProperty.VERSION] == version &&
+                    event.properties[TelemetryProperty.ATLAS_HOST] == ""
```
alenakhineika, line 72: What does this test? Why `event.properties[TelemetryProperty.ATLAS_HOST] == ""` expects always empty string? Shouldn't it be `atlastHost` like e.g. `version`.

kmruiz: This is an integration test that connects to a local cluster in Atlas and to a local cluster created with the Atlas CLI and checks the different inferred properties, that's why the ATLAS_HOST is always empty.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbe.kt
alenakhineika: It looks like we only check host URLs `val isAtlas = connectionString.hosts.all { it.matches(atlasRegex) }`, when compass also checks for atlasVersion https://github.com/mongodb-js/compass/blob/main/packages/data-service/src/instance-detail-helper.ts#L149

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
alenakhineika: Got it! Then only the same question as previously if we could make it null instead of empty string to match telemetry across all tools.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbe.kt
kmruiz: Oh, it seems that Compass is doing some additional checks. I've based the logic of this class on: https://github.com/mongodb-js/mongodb-build-info/blob/main/index.js#L5.

Do you know why Compass might do something extra?

alenakhineika: Discussed in Slack. Here is a short summary for reference.

> The atlasVersion check is the right way to identify atlas, but not all deployments return a version, so as a fallback we use `mongodb-build-info`.

The follow-up ticket [INTELLIJ-37](https://jira.mongodb.org/browse/INTELLIJ-37)

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/TelemetryServiceTest.kt
```diff
@@ -53,6 +53,7 @@ internal class TelemetryServiceTest {
                 nonGenuineServerName = null,
                 serverOsFamily = null,
                 isLocalAtlas = false,
+                atlasHost = "mongodb+srv://example-cluster.mongodb.net",
```
alenakhineika, line 56: Atlas host should be just hostname without a prefix, e.g. `example-cluster.mongodb.net`

kmruiz: It's mock data, but yeah, just to be consistent I'll remove the prefix.

alenakhineika: Yeah, this will just be helpful to others who reads the code. I usually use tests as a learning tool 📚

alenakhineika (approved):

PR #29, opened 2024-08-12
fix: fixes the toolbar showing up only on one file when multiple files are opened simultaneously INTELLIJ-57

himanshusinghs:
### Checklist

### DIFF
--- a/.github/workflows/quality-check.yaml
+++ b/.github/workflows/quality-check.yaml
@@ -100,7 +100,7 @@ jobs:
           cache: 'gradle'
       - name: Run Linter
         run: |
-          ./gradlew --quiet --console=plain spotlessApply
+          ./gradlew --quiet --console=plain spotlessCheck
 
   unit-tests:
     name: 'Unit Tests'
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,6 +6,7 @@ group = "com.mongodb"
 version = "0.0.1"
 
 plugins {
+    alias(libs.plugins.intellij) apply false
     alias(libs.plugins.versions)
     id("jacoco")
 }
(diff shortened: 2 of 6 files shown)


packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/idea/IdeaFrame.kt
```diff
@@ -51,12 +53,13 @@ class IdeaFrame(
...
-                                file
-                            ), true
-                        )
+                        const fileEditorManager = FileEditorManager.getInstance(project);
+                        const fileDescriptor = new OpenFileDescriptor(
+                            project,
+                            file
+                        );
+                        fileEditorManager.openTextEditor(fileDescriptor, true)
+                        fileEditorManager.navigateToTextEditor(fileDescriptor)
```
himanshusinghs, line 62: Added this because the opened file was not getting navigated to.

build.gradle.kts
```diff
@@ -6,6 +6,7 @@ group = "com.mongodb"
 version = "0.0.1"
 
 plugins {
+    alias(libs.plugins.intellij) apply false
```
himanshusinghs, line 9: This is because `gradlew tasks` was failing on main and a workaround was recommended [here](https://youtrack.jetbrains.com/issue/IDEA-339602/Gradle-IDEA-Ext-support-classloader-isolation).

.github/workflows/quality-check.yaml
```diff
@@ -100,7 +100,7 @@ jobs:
           cache: 'gradle'
       - name: Run Linter
         run: |
-          ./gradlew --quiet --console=plain spotlessApply
+          ./gradlew --quiet --console=plain spotlessCheck
```
himanshusinghs, line 103: Yea and this one because we probably don't wanna apply styling changes in the CI but only check?

kmruiz: Makes a lot of sense!

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/idea/IdeaFrame.kt
kmruiz: It makes sense, because the UI will open the editor when it's the only file open, but it might not do it with multiple files. Nice change!

kmruiz (approved):

PR #31, opened 2024-08-13
chore: extract collection from a model class INTELLIJ-45

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
+++ b/packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
@@ -22,14 +22,18 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.*
 import org.springframework.data.mongodb.core.MongoTemplate
 import org.springframework.data.mongodb.core.mapping.Document
+
 import java.lang.reflect.Method
 import java.net.URI
 import java.net.URL
 import java.nio.file.Paths
-import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.atomic.AtomicReference
+
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
 import kotlin.io.path.Path
 import kotlin.io.path.absolutePathString
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.suspendCancellableCoroutine
 
 @Retention(AnnotationRetention.RUNTIME)
 @Test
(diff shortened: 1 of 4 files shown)


packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
```diff
@@ -0,0 +1,210 @@
...
+
+            result.onFailure {
+                finished.set(true)
+                throwable.set(it)
+            }
+        }
+
+        while (!finished.get()) {
+            Thread.sleep(1)
+        }
```
addaleax, line 149: I guess there's no good alternative to busy-looping here?

himanshusinghs (approved):
I would love to learn more in depth about what you're doing in IntegrationTest.kt but otherwise looks good!

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private fun pathToClassJarFile(javaClass: Class<*>): String {
+        val classResource: URL =
+            javaClass.getResource(javaClass.getSimpleName() + ".class")
+                ?: throw RuntimeException("class resource is null")
+        val url: String = classResource.toString()
+        if (url.startsWith("jar:file:")) {
+            // extract 'file:......jarName.jar' part from the url string
+            val path = url.replace("^jar:(file:.*[.]jar)!/.*".toRegex(), "$1")
```
addaleax, line 196: The `!` is a literal `!` here and not some special syntax, correct?

Are we okay with this regexp potentially mis-matching if `.jar` occurs in the rest of the URL again?

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
kmruiz: Sadly we need to wait for the other thread to finish. There are other ways (maybe we can use a runBlocking coroutine). Let me take a look, it might be easier to understand and maintain.

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
kmruiz: You are right that `!` is a literal `!`, but in theory the Regex is bound to the beginning (hence the ^). However it's forbidden to have a Java class with a dot in the name so it shouldn't happen unless someone adds a malicious JAR to our testing environment. If you think I should fix it I'm fine, I think it can be easy to fix it.

addaleax: All good, that makes sense then 👍

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/IntegrationTest.kt
kmruiz: @addaleax Changed it to use a coroutine, WDYT?

addaleax: yep, that looks better! :)

addaleax (approved):

PR #32, opened 2024-08-13
chore: add dropdown to select current database when working with Spring INTELLIJ-41

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
@@ -22,7 +22,6 @@ import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiManager
 import com.intellij.psi.util.PsiModificationTracker
 import com.intellij.util.messages.MessageBusConnection
-import com.mongodb.jbplugin.accessadapter.datagrip.adapter.isMongoDbDataSource
 import com.mongodb.jbplugin.editor.MongoDbVirtualFileDataSourceProvider.Keys
 import com.mongodb.jbplugin.observability.probe.NewConnectionActivatedProbe
 import kotlinx.coroutines.CoroutineScope
@@ -64,6 +63,18 @@ class EditorToolbarDecorator(
         DaemonCodeAnalyzer.getInstance(editor.project).restart(psiFile)
     }
 
+    fun onDatabaseSelected(database: String) {
+        editor.virtualFile?.putUserData(Keys.attachedDatabase, database)
+        val psiFile = PsiManager.getInstance(editor.project!!).findFile(editor.virtualFile) ?: return
+        DaemonCodeAnalyzer.getInstance(editor.project).restart(psiFile)
+    }
+
+    fun onDatabaseUnselected() {
+        editor.virtualFile?.putUserData(Keys.attachedDatabase, null)
+        val psiFile = PsiManager.getInstance(editor.project!!).findFile(editor.virtualFile) ?: return
+        DaemonCodeAnalyzer.getInstance(editor.project).restart(psiFile)
+    }
+
     override fun selectionChanged(event: FileEditorManagerEvent) {
         (event.newEditor as? TextEditor)?.editor?.let {
             editor = it
@@ -79,8 +90,8 @@ class EditorToolbarDecorator(
             coroutineScope,
             onConnected = this::onConnected,
             onDisconnected = this::onDisconnected,
-            onDatabaseSelected = {},
-            onDatabaseUnselected = {}
+            onDatabaseSelected = this::onDatabaseSelected,
+            onDatabaseUnselected = this::onDatabaseUnselected
         )
 
         editor.project?.let { project ->
@@ -130,15 +141,30 @@ class EditorToolbarDecorator(
         }
 
         val javaPsiFile = psiFile as PsiJavaFile
-        return arrayOf(
-            this::isUsingTheJavaDriver,
-        ).any { it(javaPsiFile) }
+        if (isUsingSpringDataMongoDb(javaPsiFile)) {
+            toolbar.showDatabaseSelector()
+            return true
+        }
+
+        if (isUsingTheJavaDriver(javaPsiFile)) {
+            toolbar.hideDatabaseSelector()
+            return true
+        }
+
+        return false
     }
 
     private fun isUsingTheJavaDriver(psiFile: PsiJavaFile): Boolean {
         val importStatements = psiFile.importList?.allImportStatements ?: emptyArray()
         return importStatements.any {
             return@any it.importReference?.canonicalText?.startsWith("com.mongodb") == true
+        }
+    }
+
+    private fun isUsingSpringDataMongoDb(psiFile: PsiJavaFile): Boolean {
+        val importStatements = psiFile.importList?.allImportStatements ?: emptyArray()
+        return importStatements.any {
+            return@any it.importReference?.canonicalText?.startsWith("org.springframework.data.mongodb") == true
         }
     }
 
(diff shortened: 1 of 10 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
```diff
@@ -169,9 +139,17 @@ class EditorToolbarDecorator(
         }
 
         val javaPsiFile = psiFile as PsiJavaFile
-        return arrayOf(
-            this::isUsingTheJavaDriver,
-        ).any { it(javaPsiFile) }
+        if (isUsingSpringDataMongoDb(javaPsiFile)) {
+            toolbar.showDatabaseSelector()
```
himanshusinghs, line 143: These show and hide actions feels like a sideeffect in this method, would they be better suited when done in the `ensureToolbarIsVisibleIfNecessary`, right after we assign the toolbar to editor.permanentHeaderComponent?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
```diff
@@ -44,71 +39,38 @@ class EditorToolbarDecorator(
...
+            analyzeFileFromScratch()
         }
     }
 
-    fun onDataSourceUnselected() {
+    fun onDisconnected() {
         editor.virtualFile?.removeUserData(Keys.attachedDataSource) ?: return
-        val psiFile = PsiManager.getInstance(editor.project!!).findFile(editor.virtualFile) ?: return
-        DaemonCodeAnalyzer.getInstance(editor.project).restart(psiFile)
+        editor.virtualFile?.removeUserData(Keys.attachedDatabase)
```
himanshusinghs, line 62: We should probably remove both the keys and then return if applicable?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/MdbJavaEditorToolbar.kt
```diff
@@ -1,164 +1,181 @@
...
-                )
+        dropdowns.layout = BoxLayout(dropdowns, BoxLayout.X_AXIS)
+
+        dropdowns.add(dataSourceComboBox)
+        dropdowns.add(databaseComboBox)
+        add(dropdowns, BorderLayout.EAST)
+    }
+
+    fun showDatabaseSelector() {
+        dropdowns.add(databaseComboBox)
```
himanshusinghs, line 72: Was wondering if the `.add` call here acts like its working on a Set?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/MdbJavaEditorToolbar.kt
```diff
@@ -1,164 +1,181 @@
...
+        dataSourceComboBox.dataSources = dataSources.filter { it.isMongoDbDataSource() }
+    }
+
+    fun reloadDatabases() {
+        dataSourceComboBox.selectedDataSource?.let {
+val readModel = project.getService(DataGripBasedReadModelProvider::class.java)
+val databases = readModel.slice(dataSourceComboBox.selectedDataSource!!, ListDatabases.Slice)
+databaseComboBox.databases = databases.databases.map { it.name }
+} ?: run {
+databaseComboBox.databases = emptyList()
```
himanshusinghs, line 137: I think we need to check our Spotless config. I noticed that when we don't apply the formatting ourselves, it screws up on the indentation.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
```diff
@@ -30,7 +30,16 @@ class MdbJavaEditorToolbarFixture(
         remoteComponent,
     ) {
     val dataSources: ComboBoxFixture
-        get() = find(byXpath("//div[@class='ComboBox']"))
+        get() = find(byXpath("//div[@class='DataSourceComboBox']"))
```
himanshusinghs, line 33: Was having a concrete XPath also a motivation to refactor the MDBJavaEditorToolbar into separate comboboxes?

himanshusinghs (approved):
Great refactor <3 Learning a lot - left a few questions.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
kmruiz: The issue here is that showing the database selector depends on the driver the developer is using. We have 3 different scenarios:

* Not using MongoDB -> do not show the toolbar
* Using the Java Driver -> show the toolbar with the data source selector, but hide the database selector
* Using Spring Criteria -> show the toolbar with both the data source and database selectors

Moving this logic upwards would mean that we need to either reparse the file to check which driver is using, or returning some value (an enum maybe?) that we would use to decide what to show.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
kmruiz: Ah true, fixing it!

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
himanshusinghs: yea I was kinda implying that but if it feels a little too much then we can also just rename the method to signify that "it checks wether its a java file and perform a side-effect"

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/MdbJavaEditorToolbar.kt
kmruiz: That sounds good, I'll create a ticket for this and we can spend some time at some point. For now I'll reformat the file and push the changes. Maybe a good change is to replace the gitHook to do spotlessCheck instead of spotlessApply.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
kmruiz: To be fair no, the MDBJavaEditorToolbar was already pretty big with one single dropdown implementation and I thought that refactoring was going to make easier to add new actions later. Accessing by XPath simplifies a lot testing but I usually don't write the XPath myself :D I go to the IntelliJ inspector and take it from there.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
kmruiz: Changed to `isToolbarSetUpForCurrentFile` which seems more accurate.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
himanshusinghs: Understood - thank you for the context Kevin 🙌

PR #33, opened 2024-08-15
feat: adds inspection for field and value type mismatch INTELLIJ-23

himanshusinghs:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
@@ -83,16 +83,16 @@ class EditorToolbarDecorator(
     override fun editorCreated(event: EditorFactoryEvent) {
         editor = event.editor
 
-        toolbar = MdbJavaEditorToolbar(
-            editor.project!!,
-            coroutineScope,
-            onConnected = this::onConnected,
-            onDisconnected = this::onDisconnected,
-            onDatabaseSelected = this::onDatabaseSelected,
-            onDatabaseUnselected = this::onDatabaseUnselected
-        )
-
         editor.project?.let { project ->
+            toolbar = MdbJavaEditorToolbar(
+                project,
+                coroutineScope,
+                onConnected = this::onConnected,
+                onDisconnected = this::onDisconnected,
+                onDatabaseSelected = this::onDatabaseSelected,
+                onDatabaseUnselected = this::onDatabaseUnselected
+            )
+
             messageBusConnection = project.messageBus.connect()
             messageBusConnection.subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this)
             messageBusConnection.subscribe(DataSourceManager.TOPIC, this)
@@ -101,9 +101,9 @@ class EditorToolbarDecorator(
             messageBusConnection.subscribe(PsiModificationTracker.TOPIC, this)
             val localDataSourceManager = DataSourceManager.byDataSource(project, LocalDataSource::class.java) ?: return
             toolbar.reloadDataSources(localDataSourceManager.dataSources)
-        }
 
-        ensureToolbarIsVisibleIfNecessary()
+            ensureToolbarIsVisibleIfNecessary()
+        }
     }
 
     override fun editorReleased(event: EditorFactoryEvent) {
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
@@ -3,10 +3,8 @@ package com.mongodb.jbplugin.editor
 import com.intellij.codeInsight.daemon.DaemonCodeAnalyzer
 import com.intellij.database.console.JdbcDriverManager
 import com.intellij.database.console.session.DatabaseSessionManager
-import com.intellij.database.dataSource.DatabaseConnectionManager
 import com.intellij.database.dataSource.LocalDataSource
 import com.intellij.database.dataSource.LocalDataSourceManager
-import com.intellij.database.dataSource.connection.ConnectionRequestor
 import com.intellij.database.model.RawDataSource
 import com.intellij.database.psi.DataSourceManager
 import com.intellij.database.run.ConsoleRunConfiguration
@@ -19,14 +17,11 @@ import com.intellij.openapi.editor.ex.EditorEx
 import com.intellij.openapi.fileEditor.FileEditorManagerEvent
 import com.intellij.openapi.fileEditor.FileEditorManagerListener
 import com.intellij.openapi.fileEditor.TextEditor
-import com.intellij.openapi.rd.util.launchChildBackground
 import com.intellij.openapi.util.removeUserData
 import com.intellij.psi.PsiJavaFile
 import com.intellij.psi.PsiManager
 import com.intellij.psi.util.PsiModificationTracker
 import com.intellij.util.messages.MessageBusConnection
-import com.mongodb.jbplugin.accessadapter.datagrip.adapter.isConnected
-import com.mongodb.jbplugin.accessadapter.datagrip.adapter.isMongoDbDataSource
 import com.mongodb.jbplugin.editor.MongoDbVirtualFileDataSourceProvider.Keys
 import com.mongodb.jbplugin.observability.probe.NewConnectionActivatedProbe
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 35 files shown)


packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/BsonType.kt
```diff
@@ -15,7 +15,14 @@ import java.util.*
 /**
  * Represents any of the valid BSON types.
  */
-interface BsonType
+interface BsonType {
+    fun isMatching(otherType: BsonType): Boolean = when (otherType) {
+        this -> true
+        is BsonAny -> true
+        is BsonAnyOf -> otherType.types.any { this.isMatching(it) }
+        else -> false
```
kmruiz, line 23: I think the logic that is missing here is when `this is BsonAnyOf` something. I don't see the implementation in this fil.

kmruiz: Seen it below, the diff was a bit weird :D.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/HasValueReference.kt
```diff
@@ -18,7 +18,7 @@ data class HasValueReference(
      * @property type
      */
     data class Constant(
-        val value: Any,
+        val value: Any?,
```
kmruiz, line 21: Is this nullable because a Constant can be null? Or is it something else? Ideally Constant values can always be resolved, if they can not, we would use a Runtime value.

himanshusinghs: yea exactly because the constant itself can be null. There was a small parsing error that [fix for nullable references getting detected as BsonAny instead of Bs…](https://github.com/mongodb-js/intellij/pull/33/commits/c698b7249e18ff75e6cc3790ea5cc73bc16ce1e0) fixes and it was related to constant null being parsed as BsonAny

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
```diff
@@ -83,16 +83,16 @@ class EditorToolbarDecorator(
...
-
         editor.project?.let { project ->
+            toolbar = MdbJavaEditorToolbar(
+                project,
+                coroutineScope,
+                onConnected = this::onConnected,
+                onDisconnected = this::onDisconnected,
+                onDatabaseSelected = this::onDatabaseSelected,
+                onDatabaseUnselected = this::onDatabaseUnselected
+            )
```
himanshusinghs, line 94: This I moved inside only because I noticed that earlier we were asserting that editor.project is not null and I saw NullPointerException a couple of times in my IDE. Still not sure if it was because of this (probably not) but this change felt harmless. Let me know though if its not.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -183,7 +183,7 @@ internal class DataGripMongoDbDriver(
                 } else {
                     val decoderContext = DecoderContext.builder().build()
                     val outputCodec = codecRegistry.get(resultClass.java)
-                    val gson = Gson()
+                    val gson = GsonBuilder().serializeNulls().create()
```
himanshusinghs, line 186: This is also a drive by, just to make sure we get fields with nulls as well in our parsed schema.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
kmruiz: That sounds good! It might throw an exception if you are opening a file not in a project (like a scratch file or something) so this makes sure doesn't happen.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: Do we need null fields here? I'm fine keeping this for safety, but we should be already supporting partial documents, because this will deserialize fields that come `null` from MongoDB, but fields that are not in the document (most common case) won't be deserialized.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
```diff
@@ -63,17 +63,34 @@ abstract class AbstractMongoDbInspectionBridge(
...
+
+/**
+ * Checks whether a provided problem description has already been registered with the ProblemsHolder for a given
+ * PsiElement
+ *
+ * @param problem - Description of the problem
+ * @param source - PsiElement for which the problem is to be checked
+ * @return Boolean
+ */
+fun ProblemsHolder.isProblemAlreadyRegistered(problem: String, source: PsiElement): Boolean = this.results.any {
```
kmruiz, line 94: I'm fine keeping it for now but we should have a ticket to avoid doing this. If the issue is on the bridge, that for some reason is called twice, we will need a ticket to investigate why IntelliJ is doing this.

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/PsiMdbTreeUtil.kt
```diff
@@ -270,24 +270,30 @@ fun PsiElement.findMongoDbCollectionReference(): PsiExpression? {
 /**
  * Resolves to the value of the expression if it can be known at compile time
  * or null if it can only be known at runtime.
```
kmruiz, line 272: We should change the JavaDoc here to avoid misleading usage. I'm thinking something along the lines:

```suggestion
 * Resolves to a pair of the resolved value in the expression and whether it was possible to
 * resolve the value or not.
```

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/PsiMdbTreeUtil.kt
```diff
@@ -270,24 +270,30 @@ fun PsiElement.findMongoDbCollectionReference(): PsiExpression? {
...
  */
-fun PsiElement.tryToResolveAsConstant(): Any? {
+fun PsiElement.tryToResolveAsConstant(): Pair<Boolean, Any?> {
     if (this is PsiReferenceExpression) {
         val varRef = this.resolve()!!
         return varRef.tryToResolveAsConstant()
-    } else if (this is PsiLocalVariable) {
-        return this.initializer?.tryToResolveAsConstant()
+    } else if (this is PsiLocalVariable && this.initializer != null) {
+        this.initializer!!.tryToResolveAsConstant()
```
kmruiz, line 283: Watch out here, this should return, right? If we don't add a return here it will return false to null and ignore the expression in line 283.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
himanshusinghs: yea this was only for the fields that were explicitly null in the document. Earlier, those fields were not resulting in the end schema.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
himanshusinghs: Yea definitely - I will make a ticket for this and add a todo to remove this if unnecessary - linking to that ticket

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/PsiMdbTreeUtil.kt
himanshusinghs: oh no 😬 yea I will make this change, thank you for catching 🙌

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
himanshusinghs: Created the following ticket for investigation - https://jira.mongodb.org/browse/INTELLIJ-60 and updated the doc with the ticket as well.

kmruiz (approved):

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
kmruiz: I think it's fine, but if we have a sample of 50 documents where the value is just `null`, how will the type look like? It will be BsonAny | BsonNull?

PR #39, opened 2024-08-26
feat: Field existence and type check for SpringCriteria INTELLIJ-51, INTELLIJ-53

himanshusinghs:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
```diff
@@ -28,6 +30,11 @@ abstract class AbstractMongoDbInspectionBridge(
             "QueryForDialect${dialect.javaClass.name}",
         )
 
+    /**
+     * Ktlint complains about buildVisitor being longer than 50 lines but because it is just an object implementation
+     * it should be fine to keep it like this to favor readability
+     */
+    @Suppress("ktlint")
```
kmruiz, line 37: I agree that suppressing here is fine. However, I would suggest suppressing only the issue itself, in this case, @Suppress("TOO_LONG_FUNCTION").

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
```diff
@@ -65,21 +72,34 @@ abstract class AbstractMongoDbInspectionBridge(
...
                         }
                     }
                 }
             }
+
+            private fun queryWithCollectionReference(query: Node<PsiElement>, virtualFile: VirtualFile) =
+                MongoDbVirtualFileDataSourceProvider().getDatabase(
+                    virtualFile,
+                )?.let { database ->
+                    query.copy { component ->
```
kmruiz, line 97: Looks good! I would however put this in the model, in a method, that encapsulates this logic, and then call the method here. A method like overwriteQueryTargetDatabase or by the likes.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
himanshusinghs: oo nice - I did not know that, thank you <3

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
himanshusinghs: Actually part of it is what the query.copy is doing. I was trying to keep that a little generic for the purpose of aligning with the definition of copy method. Or do you mean to have this method (queryWithCollectionReference) entirely on the Node class?

himanshusinghs: I tried something in [0276ed0](https://github.com/mongodb-js/intellij/pull/39/commits/0276ed0291e36242dbb6f65700db2701a90acfde). Was trying to keep the method pure hence keeping it static and still doing the copy

kmruiz:
Don't forget to add the feature to the changelog :D

kmruiz (approved):

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/bridge/FieldExistenceCheckInspectionBridge.kt
himanshusinghs, line 1: This looks deleted but I just renamed this file to JavaDriverFieldExistenceCheckInspectionBridge.kt

PR #44, opened 2024-09-03
fix(toolbar): toolbar will now persist selected dataSource and database across intellij sessions

himanshusinghs:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
@@ -11,25 +11,27 @@ import com.intellij.database.run.ConsoleRunConfiguration
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.editor.Editor
-import com.intellij.openapi.editor.event.EditorFactoryEvent
-import com.intellij.openapi.editor.event.EditorFactoryListener
 import com.intellij.openapi.editor.ex.EditorEx
+import com.intellij.openapi.fileEditor.FileEditorManager
 import com.intellij.openapi.fileEditor.FileEditorManagerEvent
 import com.intellij.openapi.fileEditor.FileEditorManagerListener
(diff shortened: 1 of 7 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
```diff
@@ -47,18 +49,22 @@ private val allDialects = listOf(
...
-) : EditorFactoryListener,
+) : ProjectActivity,
     FileEditorManagerListener,
     DataSourceManager.Listener,
     JdbcDriverManager.Listener,
     PsiModificationTracker.Listener {
-    internal var inferredDatabase: String? = null
-    internal var guessedDialect: Dialect<PsiElement, Project>? = null
+    private var inferredDatabase: String? = null
+    private var guessedDialects: WeakHashMap<Editor, Dialect<PsiElement, Project>?> = WeakHashMap()
```
kmruiz, line 58: Is it safe to use the editor as the key of a map?  In editor we have the `putUserData` and `getUserData` methods, in the same way we have them in the PsiFile. Maybe we can use that to store editor-specific data? Feels more safe, we can't trust the editor to maintain the identity.

kmruiz: Also, we store the guessed dialect in the File already. Maybe we can get the dialect from there?

himanshusinghs: So that's why I created this as a WeakHashMap. The idea was when the editor instance is released the key is automatically destroyed. But I noticed that the editor instances are never released :D
And now I like your suggestion better - to store editor-specific-data on editor itself. Given that editors are not released as often, it sounds much reliable.

himanshusinghs: I will incorporate the suggestion here but first I will put through a small refactor here.

himanshusinghs:
Closing this in favor of work done in #50

PR #48, opened 2024-09-06
feat: Run Java queries directly from code INTELLIJ-70

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/EditorToolbarDecorator.kt
```diff
@@ -151,8 +151,10 @@ class EditorToolbarDecorator(
             (editor as? EditorEx)?.permanentHeaderComponent = toolbar
             editor.headerComponent = toolbar
         } ?: run {
-            (editor as? EditorEx)?.permanentHeaderComponent = null
-            editor.headerComponent = null
+            if (editor.headerComponent is MdbJavaEditorToolbar) {
```
kmruiz, line 154: This runs on every opened editor. This means that if our decorator runs after DataGrip's, we can remove their toolbar. It also can happen with the markdown editor. What we are doing here is, only remove the toolbar _if_ the toolbar that was there is ours.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/codeActions/impl/RunQueryCodeActionBridge.kt
```diff
@@ -0,0 +1,70 @@
...
+            query.sourceForMarker.textRange,
+            Icons.runQueryGutter,
+            { CodeActionsMessages.message("code.action.run.query") },
+            { _, _ ->
+                coroutineScope.launchChildBackground {
+                    val formattedQuery = MongoshDialect.formatter.formatQuery(query, explain = false)
+                    coroutineScope.launchChildOnUi {
+                        val editor = DatagripConsoleEditor.openConsoleForDataSource(query.source.project, dataSource)
+                        editor?.appendText(formattedQuery)
+                    }
```
himanshusinghs, line 63: Note to myself to ask Kevin: Would it make any difference if we were to use the context switching to switch on the EDT dispatcher for the underlying work?

PR #49, opened 2024-09-09
feat: generate index script template for queries not covered by an index INTELLIJ-74

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
@@ -19,6 +19,12 @@ import io.github.z4kn4fein.semver.Version
 import kotlinx.coroutines.CoroutineScope
 
 /**
+ * Attempts to parse a query for a PsiElement if there is one, given the dialect of the current
+ * file. Caches the result inside the PsiElement metadata and only reparses the query if that
+ * element or it's children change.
+ *
+ * It might return null if there is no query to parse.
+ *
  * @property coroutineScope
  */
 @Service(Service.Level.PROJECT)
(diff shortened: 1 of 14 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
```diff
@@ -0,0 +1,83 @@
...
+ * Attempts to parse a query for a PsiElement if there is one, given the dialect of the current
+ * file. Caches the result inside the PsiElement metadata and only reparses the query if that
+ * element or it's children change.
+ *
+ * It might return null if there is no query to parse.
+ *
+ * @property coroutineScope
+ */
+@Service(Service.Level.PROJECT)
+class CachedQueryService(
```
kmruiz, line 31: We had duplicated code across the bridges, and each bridge was parsing the query once and then caching it, that is a waste of resources. Now the cached query service is able to reuse queries parsed across different bridges and reuse them.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/IndexAnalyzer.kt
```diff
@@ -0,0 +1,93 @@
...
+            ))
+        } else {
+            otherRefs
+        }
+    }
+
+    /**
+ * @param S
+ */
+sealed interface SuggestedIndex<S> {
```
kmruiz, line 70: This can seem overkill, but I thought it would be good to have a sealed interface later, in case we support other types of indexes, like Atlas Search Indexes, that have a completely different structure.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/DataGripBasedReadModelProvider.kt
```diff
@@ -53,7 +53,7 @@ class DataGripBasedReadModelProvider(
         slice: Slice<T>,
     ): T =
         cachedValues
-            .computeIfAbsent(slice.id, fromSlice(dataSource, slice))
+            .computeIfAbsent("${dataSource.uniqueId}/${slice.id}", fromSlice(dataSource, slice))
```
himanshusinghs, line 56: Good catch 🦅

PR #51, opened 2024-09-10
fix: concurrent access to mongosh is disallowed INTELLIJ-89

kmruiz:
### Checklist


packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -257,25 +257,6 @@ internal class DataGripMongoDbDriver(
     }
 }
 
-/**
```
kmruiz, line 260: It seems we are not using this anymore, so I'm removing it.

himanshusinghs (approved):

PR #52, opened 2024-09-12
feat: query generation to run java queries in datagrip's console INTELLIJ-73

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
+++ b/packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
@@ -20,12 +20,8 @@ import com.mongodb.jbplugin.accessadapter.MongoDbDriver
 import com.mongodb.jbplugin.dialects.mongosh.MongoshDialect
 import com.mongodb.jbplugin.mql.Namespace
 import com.mongodb.jbplugin.mql.Node
-import org.bson.BsonReader
-import org.bson.BsonWriter
 import org.bson.Document
-import org.bson.codecs.Codec
 import org.bson.codecs.DecoderContext
-import org.bson.codecs.EncoderContext
 import org.bson.codecs.configuration.CodecRegistries.fromRegistries
 import org.bson.codecs.configuration.CodecRegistry
 import org.bson.conversions.Bson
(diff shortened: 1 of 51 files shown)


packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/adapter/DataGripMongoDbDriver.kt
```diff
@@ -81,7 +84,27 @@ internal class DataGripMongoDbDriver(
 
     override suspend fun connectionString(): ConnectionString = ConnectionString(dataSource.url!!)
 
-    override suspend fun <S> explain(query: Node<S>): ExplainPlan = ExplainPlan.CollectionScan
+    override suspend fun <S> explain(query: Node<S>): ExplainPlan = withContext(Dispatchers.IO) {
```
kmruiz, line 87: This connects the explain plan to the query generation. It's the work we need to do for INTELLIJ-93, but it was very convenient to work on it so we could test that the generated query was running on the index from within the IDE, and it wasn't a lot of work.

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/MongoshDialectFormatter.kt
```diff
@@ -1,16 +1,27 @@
...
+            emitQueryBody(query, firstCall = true)
+        })
+        if (explain) {
+            emitPropertyAccess()
+            emitFunctionName("explain")
+            emitFunctionCall()
+        }
+    }.computeOutput()
+
     override fun <S> indexCommandForQuery(query: Node<S>): String = when (val index = IndexAnalyzer.analyze(query)) {
```
kmruiz, line 25: At some point we might want to improve the MongoshBackend to also generate the index creation script through the backend itself, instead of templating. However, it's not something really urgent at all so we can just do it whenever the index creation script gets more complicated.

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/backend/MongoshBackend.kt
```diff
@@ -0,0 +1,187 @@
...
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.util.*
+
+private const val MONGODB_FIRST_RELEASE = "2009-02-11T18:00:00.000Z"
+
+/**
+ * @param context
+ */
+class MongoshBackend(private val context: Context = DefaultContext()) : Context by context {
```
kmruiz, line 18: Context by context means that MongoshBackend implements Context, and delegates all the methods of Context to the variable context. It's done like this for convenience because we need to register variables and constants into the backend from within the mongosh formatter

GaurabAryal (approved):

PR #53, opened 2024-09-12
fix: for persisting datastore selection for a workspace between IntelliJ sessions INTELLIJ-83

himanshusinghs:
## Description
Came across this while testing INTELLIJ-83. We were persisting settings per application level and not per project. This PR fixes that and now we will be persisting settings per project level

### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/MdbPluginDisposable.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/MdbPluginDisposable.kt
@@ -1,6 +1,8 @@
 package com.mongodb.jbplugin.editor.services
 
+import com.intellij.configurationStore.StoreUtil
 import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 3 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/MdbPluginDisposable.kt
```diff
@@ -33,6 +35,12 @@ class MdbPluginDisposable : Disposable {
      * Disposable interface that gets called when this service is being disposed
      */
     override fun dispose() {
+        // At the very least we would always want to persist our modified settings (ToolbarSettings, PluginSettings)
+        // when closing the project which is why we explicitly invoke StoreUtil.saveSettings to persist them to disk
+        ApplicationManager.getApplication().invokeLater {
+            StoreUtil.saveSettings(ApplicationManager.getApplication())
+        }
```
himanshusinghs, line 42: This fixes the problem of ToolbarSettings not getting persisted when a project was closed or application restarted right after making a selection

kmruiz (approved):

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/MdbPluginDisposable.kt
kmruiz: Now that it's a project settings, shouldn't be StoreUtil.saveSettings(project)?

himanshusinghs: it turns out I don't need this anymore - I added this before converting the ToolbarSettings to project level service and now that it is a project level service, it gets disposed off correctly when closing the project.

PR #55, opened 2024-09-16
chore: instead of running the whole pipeline again, only allow drafts when main is green already

kmruiz:
### Checklist

### DIFF
--- a/.github/workflows/draft-release.yaml
+++ b/.github/workflows/draft-release.yaml
@@ -68,47 +68,10 @@ jobs:
             echo "If this version has already been release consider using a different one."
             exit 1
           fi
-      - name: Setup Docker (for Test Containers)
+      - name: Verify Checks in main
         run: |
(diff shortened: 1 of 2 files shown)


himanshusinghs:
I think its a good idea but there is one thing this might miss, last minute dependency problem creeps (we saw that being quite common in compass repo). Something we should be aware of.

kmruiz:
We are using fixed versions, it shouldn't happen right?

himanshusinghs:
Yea but it might happen that our fixed version was identified for a vuln at some point, right? :D

kmruiz:
Ah true, this might happen for sure. So what we can do is, whenever we generate the SBOM (we are not doing it yet but it's planned), do it also before releasing and after checking main is green. So we don't need to run the whole suite but we verify that our deps don't have any important CVE.

himanshusinghs (approved):

PR #56, opened 2024-09-18
chore: run query gutter icon will always show and allow to select a connection INTELLIJ-92

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
+++ b/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
@@ -53,12 +53,12 @@ fun eventually(timeout: Duration = Duration.ofSeconds(1), fn: () -> Unit) {
  * @param fn
  * @return
  */
-fun <T> eventually(timeout: Duration = Duration.ofSeconds(1), fn: () -> T): T = waitFor<T>(timeout, Duration.ofMillis(
+fun <T> eventually(timeout: Duration = Duration.ofSeconds(1), fn: () -> T): T? = waitFor<T?>(timeout, Duration.ofMillis(
 50
 )) {
         val result = runCatching {
             fn()
         }
 
-        result.isSuccess to result.getOrThrow()
+        result.isSuccess to result.getOrNull()
     }
--- a/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
+++ b/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
@@ -33,3 +33,32 @@ fun eventually(timeout: Duration = Duration.ofSeconds(1), fn: () -> Unit) {
         }.getOrDefault(false)
     }
 }
+
+/**
+ * Waits until the block function finishes successfully up to 1 second (or the provided timeout).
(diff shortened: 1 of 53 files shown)


packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
```diff
@@ -4,9 +4,23 @@
...
+ * Closes all open modals by using the ESC key. This should be enough for recovering on a lot of different test cases.
+ */
+fun RemoteRobot.closeAllOpenModals() {
+    // just do it a few times, to ensure we closed all modals
+    for (i in 0..5) {
+        keyboard {
+            escape()
+        }
+    }
+}
```
himanshusinghs, line 22: nit: It might serve us better to rename this to `closeNOpenModals(n: Number)`?

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/UiTestExtensions.kt
```diff
@@ -65,6 +71,10 @@ private class UiTestExtension :
     ): Any = remoteRobot
 
     override fun beforeAll(context: ExtensionContext?) {
+        if (!StepWorker.processors.containsElements { it is StepLogger }) {
+            StepWorker.registerProcessor(StepLogger())
+        }
```
himanshusinghs, line 76: What is a StepWorker? and what is this piece doing?

himanshusinghs (approved):
Amazing work @kmruiz 🚀

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/UiTestExtensions.kt
himanshusinghs: Also is it somehow related to the step calls you are doing in [IdeaFrame.kt](https://github.com/mongodb-js/intellij/pull/56/files#diff-0f6b1cf984fc49ad87c70eec12f076ccc20b131be0c6e016278bc40bbeb28869)

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
kmruiz: I wanted to keep the number hidden because the purpose is like "cleaning the modals". 5 is a magic number, hopefully it works for our use cases. Usually you won't have more than 1 or 2 modals open.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/UiTestExtensions.kt
kmruiz: ah, this is for logs. This will show all the `steps` that we've done in the test in the logs.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
himanshusinghs: but this should not worry us - its single place so we can surely fix easily when something is not working

PR #60, opened 2024-09-24
fix: BsonType not assignable to BsonArray INTELLIJ-94

himanshusinghs:
## Description
In MQL it is permissible to have a query like { countries: "Germany" } where countries is a list of string in mongodb schema. We were incorrectly marking such assignments as a possible flaw.

This commit fixes it by making sure that a given BsonType is assignable to a BsonArray as well as long as the given BsonType is also assignable to the underlying schema of the BsonArray

### Checklist


kmruiz:
The code looks good, but it still warns about nullable values, also inside arrays. Do you want to fix it here? The use case is:

I have a nullable value in Java (an Integer for example) and in MongoDB, I have a column that is an int and is not nullable. This happens because IIRC, Integer is mapped to AnyOf(BsonInt32, BsonNull) and this is not compatible right now with BsonInt32 (because of the null).

kmruiz (approved):

PR #62, opened 2024-09-26
chore: support for not in query generation INTELLIJ-77

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/MongoshDialectFormatter.kt
+++ b/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/MongoshDialectFormatter.kt
@@ -118,15 +118,19 @@ private fun <S> MongoshBackend.emitQueryBody(
                 if (firstCall) {
                     emitObjectStart()
                 }
-                if (fieldRef != null && valueRef != null) {
+                if (fieldRef != null) {
                     emitObjectKey(resolveFieldReference(fieldRef))
-                    emitContextValue(resolveValueReference(valueRef, fieldRef))
-                } else {
-                    hasChildren?.children?.forEach {
-                        emitQueryBody(it)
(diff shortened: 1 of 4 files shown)


packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/MongoshDialectFormatter.kt
```diff
@@ -172,6 +177,57 @@ private fun <S> MongoshBackend.emitQueryBody(
...
+                val operation = innerChild.component<Named>()
+                val fieldRef = innerChild.component<HasFieldReference<S>>()
+                val valueRef = innerChild.component<HasValueReference<S>>()
+
+                if (fieldRef == null) { // we are in an "and" / "or"...
+                    // so we use $nor instead
+                    emitQueryBody(
+                        Node(
+                            node.source,
+                            node.components<Component>().filterNot { it is Named } + Named(Name.NOR)
```
himanshusinghs, line 201: Why do we emit Nor when it could also be inside an And?

kmruiz: Because when you try to run a $not / $and the mongosh fails and suggests to use $nor for some reason, so I'm applying this automatically.

himanshusinghs: Interesting - I was not aware of this behavior. Its also in the documentation:
```
[$nor](https://www.mongodb.com/docs/manual/reference/operator/query/nor/#mongodb-query-op.-nor) performs a logical NOR operation on an array of one or more query predicates and selects the documents that fail all the query predicates in the array
```
Thank you for the context :)

himanshusinghs (approved):

PR #68, opened 2024-10-09
Poc/di

kmruiz:
This PR plays a bit with Kotlin delegates to simplify dependency injection. I would like to have some more compile time checks to be fair, but this kind of works fine.

1. Implements a new delegate: `by service` that loads an application service. Can be used in two ways thanks to Kotlin type inference.

```kt
val telemetry: TelemetryService by service()
val telemetry by service<TelemetryService>()
```

2. Implements a new delegate: `by Project.service` that works the same way:

```kt
val project: Project
val myService by project.service<MyService>()
```

3. Implements granular setting access, to make it easier to work with settings.
```kt
val isTelemetryEnabled by pluginSetting { ::isTelemetryEnabled }
```

4. Implements also a `SharedFlow.latest` delegate that computes in background the property on each event received on the shared flow transparently for the user. This is useful to propagate UI changes:

```kt
class EventNumber(val number)
val sharedFlow: SharedFlow<EventNumber>

val latestNumber by sharedFlow.latest { number }
val doubleNumber by sharedFlow.latest { number * 2 }
```
```

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
@@ -9,15 +9,15 @@ import com.intellij.notification.NotificationGroupManager
 import com.intellij.notification.NotificationType
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
-import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.options.ShowSettingsUtil
 import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.startup.StartupActivity
 import com.mongodb.jbplugin.i18n.TelemetryMessages
+import com.mongodb.jbplugin.meta.injecting
 import com.mongodb.jbplugin.observability.probe.PluginActivatedProbe
 import com.mongodb.jbplugin.settings.PluginSettingsConfigurable
-import com.mongodb.jbplugin.settings.useSettings
+import com.mongodb.jbplugin.settings.pluginSetting
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
@@ -54,13 +54,14 @@ class ActivatePluginPostStartupActivity(
 ) : StartupActivity, DumbAware {
     override fun runActivity(project: Project) {
         cs.launch {
-            val pluginActivated = ApplicationManager.getApplication().getService(
-                PluginActivatedProbe::class.java
-            )
+            val pluginActivated by injecting<PluginActivatedProbe>()
             pluginActivated.pluginActivated()
 
-            val settings = useSettings()
-            if (!settings.hasTelemetryOptOutputNotificationBeenShown) {
+            var firstTimeTelemetry by pluginSetting<Boolean> {
+                ::hasTelemetryOptOutputNotificationBeenShown
+            }
+
+            if (!firstTimeTelemetry) {
                 NotificationGroupManager.getInstance()
                     .getNotificationGroup("com.mongodb.jbplugin.notifications.Telemetry")
                     .createNotification(
@@ -73,7 +74,7 @@ class ActivatePluginPostStartupActivity(
                     .addAction(OpenMongoDbPluginSettingsAction())
                     .notify(project)
 
-                settings.hasTelemetryOptOutputNotificationBeenShown = true
+                firstTimeTelemetry = true
             }
         }
     }
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
@@ -14,7 +14,7 @@ import com.intellij.openapi.project.DumbAware
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.startup.StartupActivity
 import com.mongodb.jbplugin.i18n.TelemetryMessages
-import com.mongodb.jbplugin.meta.injecting
+import com.mongodb.jbplugin.meta.service
 import com.mongodb.jbplugin.observability.probe.PluginActivatedProbe
 import com.mongodb.jbplugin.settings.PluginSettingsConfigurable
 import com.mongodb.jbplugin.settings.pluginSetting
@@ -54,7 +54,7 @@ class ActivatePluginPostStartupActivity(
 ) : StartupActivity, DumbAware {
     override fun runActivity(project: Project) {
         cs.launch {
-            val pluginActivated by injecting<PluginActivatedProbe>()
+            val pluginActivated by service<PluginActivatedProbe>()
             pluginActivated.pluginActivated()
 
             var firstTimeTelemetry by pluginSetting<Boolean> {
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

**Comment by kmruiz** [comment]
For the settings change, the main motivation is safety. Currently we expose the whole state, which contains mutable properties. So every time you access the settings, you have write access to them.

However, *most of the time* you only want to query data from a setting, so you want immutable access to it.

This pattern will allow us to be explicit whether we want to write or not.

`val mySetting by pluginSetting`

Immutable, uses val, won't compile.

`var mySetting by pluginSetting`

Mutable, can write into it.

**Comment by kmruiz** [comment]
For the dependency injection mechanism, the main reason is ergonomics. Delegates allow us to decouple a bit the source of the service and eventually will allow us to add compiler plugins or inspections to make sure we do proper injection.

For example, not compiling when we use service<A> and A is a project service.

**Comment by kmruiz** [comment]
Lastly, for the sharedFlow.async: Managing state in concurrent environments is hard, and we need to handle eventual consistency. This delegate allows us to use an ordinary variable that is up to date to the latest event of a shared flow, avoiding concurrent updates.

It allows us to, either use a Flux pattern where the state flow is:

Event -> Event Queue -> Derived State 

Without worrying about coroutines or multi threaded. If the shared flow is configured to replay the latest event to new subscribers, we don't even need to worry about not having an initialised state.

This enables us also to work on a `kind of` Redux flow for shared immutable state with concurrently safe updates and `selectors` that get them. However, I'm quite against having some kind of global state, I think this flow should be isolated in each component as we are doing right now. This already allowed us pretty big refactors of the toolbar w/o breaking any other module.

**Review by addaleax — COMMENTED** [review]
Cool! The code looks all good, I just lack the background knowledge to say how this aligns or doesn't align with best practices for Kotlin in general

File: packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt [hunk 1]
```diff
@@ -73,7 +74,7 @@ class ActivatePluginPostStartupActivity(
                     .addAction(OpenMongoDbPluginSettingsAction())
                     .notify(project)
 
-                settings.hasTelemetryOptOutputNotificationBeenShown = true
+                firstTimeTelemetry = true
```
himanshusinghs, line 77: Something I want to bring attention to - When I first had a look at the original code of updating the PersistentStateComponent, I was a little amazed that simply setting a variable on a class instance will persist the state, was already kind of a magic(until I read how it works). But this feels a lot more magical that simply setting a local variable is supposed to update the persistent state. I do understand now how it works and the understandable side of me thinks of this being natural now but I think for someone new this might be confusing, something we should be aware of? Let me know what you think?

himanshusinghs: Over the last few months, I developed a preference of explicit setter and getter functions over kotlin properties, helps me understand clearly what's gonna happen so for example. Imagine the above code being.
```
if (pluginSetting.firstTimeTelemetry) {
  // pluginSetting.firstTimeTelemetry is immutable
}

pluginSetting.updateAndPersist("firstTimeTelemetry", true)
```
(I understand and agree that this looks very 90'ish 😆)

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/meta/BuildInformation.kt
```diff
@@ -16,6 +16,6 @@ object BuildInformation {
     private val properties: Properties = Properties().also {
         it.load(BuildInformation::class.java.getResourceAsStream("/build.properties"))
     }
-    val pluginVersion = properties["pluginVersion"]!!.toString()
-    val segmentApiKey = properties["segmentApiKey"]!!.toString()
+    val pluginVersion: String by properties
+    val segmentApiKey: String by properties
```
himanshusinghs, line 20: wow this is cool 🤯

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/settings/PluginSettings.kt
```diff
@@ -26,23 +28,46 @@ class PluginSettingsStateComponent : SimplePersistentStateComponent<PluginSettin
...
+        value: T
+    ) {
+        settingProp.setter.call(value)
+    }
+}
+
+fun <T> pluginSetting(cb: PluginSettings.() -> KMutableProperty0<T>): SettingsDelegate<T> {
+    val settings by service<PluginSettingsStateComponent>()
+    return SettingsDelegate(settings.state.cb())
+}
```
himanshusinghs, line 73: This is really amazing. I will check Kotlin docs for more reference to this, feel free to drop any other reference that might be helpful.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/settings/PluginSettingsConfigurable.kt
```diff
@@ -27,21 +28,22 @@ class PluginSettingsConfigurable : Configurable {
...
+            savedSettings.state.isTelemetryEnabled
     }
 
     override fun apply() {
-        val savedSettings =
-            useSettings().apply {
-                isTelemetryEnabled = settingsComponent.isTelemetryEnabledCheckBox.isSelected
-            }
+        val savedSettings by service<PluginSettingsStateComponent>()
+        savedSettings.state.apply {
```
himanshusinghs, line 38: yea so this is one such side effect, that what we're exposing as service is an implementation specifics. If the service was not fetching PluginSettingsStateComponent but lets say "PluginSettings" class with interface such as `setTelemetryEnabled`, etc then we won't need to worry about digging `.state` out of savedSettings :D

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/meta/DependencyInjection.kt
```diff
@@ -0,0 +1,135 @@
...
+    onNewState: E.() -> T,
+    onChange: suspend T.() -> Unit = {}
+
+): AsyncState<E, T> =
+    AsyncState(
+        parent,
+        this,
+        onNewState,
+        onChange
+    )
```
himanshusinghs, line 135: I am gonna try to use this in ToolbarModel, its exactly what we need for those reactive UI states. Thanks for this <3

himanshusinghs (approved):
This is really cool update Kevin. In general I really like the idea, I raised one concern regarding one thing feeling a little magical but I don't think that's a blocker for this change, just something to consider maybe?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ActivatePluginPostStartupActivity.kt
kmruiz: I personally prefer method calls, but I was testing property delegates because you can tell the compiler if something is mutable or not, whether using methods is not possible. You would need like two interfaces: one that only contains getters, and only that contains both getters and setters, and I feel this would be more cumbersome.

WDYT?

himanshusinghs: It should be possible with methods as well to tell if something is mutable or not. A getter would only return an immutable value and one would use a setter for mutating that. I agree that it starts to become cumbersome when you intend to have more than a few properties and then corresponding getter/setter methods for those properties but then there we could use some generic method accepting key/value combo (not yet sure how that would work here in kotlin, so pardon me for bringing some TS concept here 😆).

I have no strong feelings against the delegate version, I was just a little surprised at first when I read in code that mutating a local variable in function also updates the state of the object from which it was retrieved from. But then I read the Java docs and it was clear how this works so I think at this point we can even go ahead and try this concept and see how it works for us?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/settings/PluginSettingsConfigurable.kt
himanshusinghs: yea that sounds good 👍

PR #69, opened 2024-10-10
fix: fix for connection state and selection state not propagating to the Toolbar INTELLIJ-96

himanshusinghs:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/MdbJavaEditorToolbar.kt
```diff
@@ -167,129 +138,91 @@ class MdbJavaEditorToolbar(
...
-                        ApplicationManager.getApplication().invokeLater {
-                            onConnectionStateChange(
-                                connectionState,
-                                toolbarState.selectedDataSource
-                            )
-                        }
-                    }
+                val toolbarModel = project.getToolbarModel()
+                val oldToolbarState = toolbarModel.toolbarState.value
+                if (!dialog.showAndGet()) {
```
kmruiz, line 169: Why does the state change when the user cancels the dialog? Is it sharing the same reference somehow?

himanshusinghs: yea so all the UI stuff now talk to this singleton ToolbarModel and they apply the changes there. In this particular case we revert the change on cancellation.

himanshusinghs:
Merged this despite UI test failures as they are consistent flakes.

PR #75, opened 2024-10-15
chore: fixes for failing unit test and minor UI bug

himanshusinghs:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
+++ b/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
@@ -83,7 +83,7 @@ internal class NewConnectionActivatedProbeTestForLocalEnvironment :
         isLocalhost = true,
         isEnterprise = false,
         isGenuine = true,
-        version = "7.0.14",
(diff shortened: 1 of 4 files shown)


packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
```diff
@@ -83,7 +83,7 @@ internal class NewConnectionActivatedProbeTestForLocalEnvironment :
         isLocalhost = true,
         isEnterprise = false,
         isGenuine = true,
-        version = "7.0.14",
+        version = "7.0.12",
```
himanshusinghs, line 86: This was unexpected, I guess they took down 7.0.14 (for whatever reason)

kmruiz (approved):
Amazing work!

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/observability/probe/NewConnectionActivatedProbeTest.kt
kmruiz: Maybe. At some point we will need to update these tests to 8.0, but we don't need to do it now.

PR #78, opened 2024-10-18
chore: improvements on flakiness of UI tests INTELLIJ-107

himanshusinghs:
## Description
Two things that I identified inducing flakiness in our UI tests were:
1. ComboBox getting updated while tests were trying to use them, made the combobox collapse
2. Tests were not able to recover from unexpected state and eventually failed

To address the first concern, we are now updating the combobox only when there is an actual change, otherwise not and that too wrapped in an invokeLater so that existing interactions are finished before performing the update.

Second concern is addressed in two steps:
1. We do not want to interact with the comboboxes, right after they show up on screen because they subscribe to the StateFlow and eventually a state update follows which leads to the disruption in UI interaction. So we have put a short timeout of 1 second to allow for initial state subscription.
2. The state transitions that depend on related are grouped together in retry-able block so that when the tests end up in an expected state they can retry from the start to achieve the expected state. One such example is `openRunQueryPopup`.

### Update
There were two further problems identified for subsequent failures:
1. Gradle project would never load for the very first start of the test run
2. Tests tearDown will fail complaining about writes happening in write safe context

To address the first problem, we have worked up a series of UI steps to always perform for every test that will ensure that gradle project is in sync.

And for the second problem, we realised that the write unsafe error is more of a warning which can safely be ignored. But to be on the correct step here, we have wrapped all the actions now in an invokeLater with explicit modality state.

Note: This PR does not entirely remove flakes from our CI. There is still one source of problem (happening from time to time) which I haven't been able to understand completely. When we open a file, at times it will happen that the toolbar is not visible. There is no gradle sync or indexing happening in the background but still its almost like the IDE is in dumb mode. Not yet entirely sure why but I am happy to look into that as a follow up.

### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
@@ -1,6 +1,7 @@
 package com.mongodb.jbplugin.editor.inputs
 
 import com.intellij.database.dataSource.LocalDataSource
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.ComboBox
 import com.intellij.openapi.ui.asSequence
@@ -21,7 +22,6 @@ import java.awt.event.ItemListener
 import javax.swing.DefaultComboBoxModel
 import javax.swing.JComponent
 import javax.swing.SwingConstants
-import javax.swing.SwingUtilities
 import javax.swing.event.PopupMenuEvent
 import javax.swing.event.PopupMenuListener
 
@@ -100,7 +100,7 @@ class DataSourceComboBox(
         var isFirstInit = true
         coroutineScope.launch {
             project.getToolbarModel().toolbarState.collect { state ->
-                SwingUtilities.invokeLater {
+                ApplicationManager.getApplication().invokeLater {
                     updateComboBoxState(state, isFirstInit)
                     isFirstInit = false
                 }
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
@@ -2,6 +2,7 @@ package com.mongodb.jbplugin.editor.inputs
 
 import com.intellij.database.dataSource.LocalDataSource
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.ModalityState
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.ui.ComboBox
 import com.intellij.openapi.ui.asSequence
@@ -100,10 +101,10 @@ class DataSourceComboBox(
         var isFirstInit = true
         coroutineScope.launch {
             project.getToolbarModel().toolbarState.collect { state ->
-                ApplicationManager.getApplication().invokeLater {
+                ApplicationManager.getApplication().invokeLater({
                     updateComboBoxState(state, isFirstInit)
                     isFirstInit = false
-                }
+                }, ModalityState.stateForComponent(comboBoxComponent))
             }
         }
     }
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
@@ -8,6 +8,7 @@ import com.intellij.sql.indexOf
 import com.intellij.ui.AnimatedIcon.ANIMATION_IN_RENDERER_ALLOWED
 import com.intellij.ui.components.JBLabel
 import com.mongodb.jbplugin.accessadapter.datagrip.adapter.isConnected
+import com.mongodb.jbplugin.editor.models.ToolbarState
 import com.mongodb.jbplugin.editor.models.getToolbarModel
 import com.mongodb.jbplugin.i18n.Icons
 import com.mongodb.jbplugin.i18n.Icons.scaledToText
@@ -20,6 +21,7 @@ import java.awt.event.ItemListener
 import javax.swing.DefaultComboBoxModel
 import javax.swing.JComponent
 import javax.swing.SwingConstants
+import javax.swing.SwingUtilities
 import javax.swing.event.PopupMenuEvent
 import javax.swing.event.PopupMenuListener
 
@@ -98,14 +100,9 @@ class DataSourceComboBox(
         var isFirstInit = true
         coroutineScope.launch {
             project.getToolbarModel().toolbarState.collect { state ->
-                withoutSelectionChangedListener {
-                    selectedDataSourceConnecting = state.selectedDataSourceConnecting
-                    selectedDataSourceFailedConnecting = state.selectedDataSourceConnectionFailed
-                    if (isFirstInit || state.dataSources != dataSources) {
-                        isFirstInit = false
-                        populateComboBoxWithDataSources(state.dataSources)
-                    }
-                    selectDataSourceByUniqueId(state.selectedDataSource?.uniqueId)
+                SwingUtilities.invokeLater {
+                    updateComboBoxState(state, isFirstInit)
+                    isFirstInit = false
                 }
             }
         }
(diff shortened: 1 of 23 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
```diff
@@ -98,14 +100,9 @@ class DataSourceComboBox(
...
             project.getToolbarModel().toolbarState.collect { state ->
-                withoutSelectionChangedListener {
-                    selectedDataSourceConnecting = state.selectedDataSourceConnecting
-                    selectedDataSourceFailedConnecting = state.selectedDataSourceConnectionFailed
-                    if (isFirstInit || state.dataSources != dataSources) {
-                        isFirstInit = false
-                        populateComboBoxWithDataSources(state.dataSources)
-                    }
-                    selectDataSourceByUniqueId(state.selectedDataSource?.uniqueId)
+                SwingUtilities.invokeLater {
```
kmruiz, line 103: Why are we using the SwingUtilities invokeLater and not the Application invokeLater?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DatabaseComboBox.kt
```diff
@@ -78,13 +80,9 @@ class DatabaseComboBox(
...
         coroutineScope.launch {
             project.getToolbarModel().toolbarState.collect { state ->
-                withoutSelectionChangedListener {
-                    loadingDatabases = state.databasesLoadingForSelectedDataSource
-                    if (isFirstInit || state.databases != databases) {
-                        isFirstInit = false
-                        populateComboBoxWithDatabases(state.databases)
-                    }
-                    selectDatabaseAndNotify(state.selectedDatabase)
+                SwingUtilities.invokeLater {
```
kmruiz, line 83: Same comment as above, is there an specific reason to use SwingUtilities instead of Application invokeLater?

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
```diff
@@ -43,11 +45,36 @@ class MdbJavaEditorToolbarFixture(
...
         }.isSuccess
-    fun detachDataSource() {
-        eventually {
-            dataSources.selectItem("Detach data source")
-            if (!dataSources.selectedText().contains("Attach MongoDB")) {
-                throw Exception("Could not detach data source")
+
+    fun selectDataSource(title: String) {
+        eventually(1.minutes.toJavaDuration()) {
+            step("Selecting DataSource $title in toolbar") {
```
kmruiz, line 51:
```suggestion
            step("Selecting DataSource $title in toolbar") {
                eventually(1.minutes.toJavaDuration()) {
```

Because eventually will retry every a few milliseconds, until (in this case) 1 minute passes, so we will see the step in the logs for every retry.

kmruiz: If we are going to use this pattern of eventually/step a lot, maybe we want to have an eventually that receives the step name and does both things? So we encapsulate both calls and it's easier to use.

kmruiz (changes requested):

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
himanshusinghs: Actually in practice I found that step log for every retries were making it more clear about what is happening. I am thinking of step as very detailed and verbose explanation of what is happening in the UI. Agree that we probably should have eventually with a step block. I will work towards that if my earlier reasoning sounds good?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
himanshusinghs: This is what my IDE autocompleted when I used invokeLater. I now see that in every other place we use Application.invokeLater, will switch to that but just for my clairification - are these two same or not?

kmruiz: Application.invokeLater does a few optimisations depending on the thread you are one and it's IntelliJ specific, so we should do that unless we have a really strong reason to avoid it.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
kmruiz: It sounds great, if it's possible can we add then the retry number? Like "Selecting DataSource $title: retry N" so it's clear in the logs that we are retrying?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/inputs/DataSourceComboBox.kt
himanshusinghs: No reason other than autocomplete - I will make the change. Thank you for the suggestion 🙌

kmruiz (approved):

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/codeActions/impl/MongoDbRunQueryActionUiTest.kt
```diff
@@ -20,12 +21,15 @@ class MongoDbRunQueryActionUiTest {
         remoteRobot: RemoteRobot,
         url: MongoDbServerUrl,
     ) {
+        // This is our safety net to ensure we start with a clean slate
+        remoteRobot.openDatabaseToolWindow().removeAllDataSources()
+        remoteRobot.closeRightToolWindow()
```
himanshusinghs, line 26: In 99.99% of cases, this will be a no-op but I am keeping this as a safety net to reduce the flakes.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/codeActions/impl/MongoDbRunQueryActionUiTest.kt
```diff
@@ -20,12 +21,15 @@ class MongoDbRunQueryActionUiTest {
...
         remoteRobot.ideaFrame().addDataSourceWithUrl(javaClass.simpleName, url)
     }
 
     @AfterEach
     fun tearDown(remoteRobot: RemoteRobot) {
-        remoteRobot.ideaFrame().cleanDataSources()
+        remoteRobot.openDatabaseToolWindow().removeAllDataSources()
```
himanshusinghs, line 32: So this is the change that makes sure that our tearDown don't fail. The cleanDataSources will fail from time to time, depending on the modality state of the application. Performing the same action via UI is working much more reliably.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
```diff
@@ -21,6 +22,38 @@ fun RemoteRobot.closeAllOpenModals() {
...
+    timeout: Duration = Duration.ofSeconds(1),
+    recovery: () -> Unit = {},
+    fn: (Int) -> Unit
+) {
+    eventually(timeout, recovery) { attempt ->
+        step("$description, attempt=$attempt") {
+            fn(attempt)
+        }
+    }
+}
```
himanshusinghs, line 55: Added this to address the suggestion about combining eventually and step descriptions.

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/components/MdbJavaEditorToolbarFixture.kt
himanshusinghs: Added this [here](https://github.com/mongodb-js/intellij/pull/78#discussion_r1812621034)

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/FixtureExtensions.kt
```diff
@@ -162,26 +161,12 @@ fun RemoteRobot.openProject(absolutePath: String) {
...
 /**
  * Closes the project and waits until properly closed.
  */
-fun RemoteRobot.closeProject() {
+fun RemoteRobot.closeProject() = step("Closing any open project") {
     invokeAction("CloseProject")
-    maybeTerminateButton()
-}
-
-private fun RemoteRobot.maybeTerminateButton() {
```
kmruiz, line 178: This was necessary when a process is running when a test finishes (like a gradle sync). It opens a pop up with the button. Why are we getting rid of this? Is not necessary anymore?

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/UiTestExtensions.kt
```diff
@@ -126,18 +129,26 @@ private class UiTestExtension :
...
 
-            if (it.smartMode) {
+            if (project.smartMode) {
                 remoteRobot.ideaFrame().disablePowerSaveMode()
+                remoteRobot.openGradleToolWindow().also {
+                    it.ensureGradleProjectsAreSynced()
+                }
+                // This ideally should not be needed anymore, but we still perform this
+                // to wait for smart mode to kick-in before doing anything else.
                 remoteRobot.ideaFrame().waitUntilProjectIsInSync()
```
kmruiz, line 147: I think it's fine to keep this. If the project is already in sync and in smart mode, it should be a noop, so I don't think it will slow the test down much.

kmruiz (approved):

PR #79, opened 2024-10-23
feat: basic support for the @Query annotation INTELLIJ-114/INTELLIJ-117

kmruiz:
## Description

We don't have a reliable way to test how `@Query` works right now due to the error we are having in our test configuration related to injected languages. Due to this, we skip INTELLIJ-115 (the actual parsing of the query) and we do 114 and 117 which are more related to metadata of the query.

### Checklist


.idea/intellij.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
```
himanshusinghs, line 2: Is this intended to be committed? If so then I wonder why is the type JAVA_MODULE?

packages/mongodb-dialects/spring-@query/src/main/kotlin/com/mongodb/jbplugin/dialects/springquery/SpringAtQueryDialectParser.kt
```diff
@@ -0,0 +1,333 @@
...
+                    return listOf(
+                        Node(
+                            source,
+                            listOf(
+                                opName,
+                                resolveValueReference(source.children[1], parent)
+                            )
+                        )
+                    )
+                } else if (isImplicitAnd(source.children[1])) {
```
himanshusinghs, line 112: I feel like we should be a bit cautious when passing indexed values because they might not exist in some cases, as we saw in the past. Unless we are not parsing when the query is incomplete?

himanshusinghs (approved):
Amazing work and also looks good.

.idea/intellij.iml
kmruiz: IDK why intellij created it, but I'm going to add it to .gitignore

PR #81, opened 2024-10-29
fix: makes a List assignable to simple type INTELLIJ-120

himanshusinghs:
When using a $in filter, it is expected to provide a List of string/number, etc to be matched against a field which itself is a simple type (string/number, etc). This PR marks such assignment acceptable so that our linter does not throw false positive warnings.

### Checklist


kmruiz (approved):
🚀

PR #92, opened 2024-11-14
chore: mql specs

kmruiz:
### Checklist


packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,145 @@
...
+* MongoDB Community 8.0, from a development environment, does not have an index on bookName for the target collection.
+* MongoDB Atlas 8.0, production environment, does have an index on bookName for the target collection.
+
+In the development environment, we copy the data from the production environment once every week. For this example,
+we will consider that the data sets are exactly the same on both clusters.
+
+| Cluster Environment           	 | Is Valid 	  | Is Supported 	 | Same Results 	 | Same Dialect 	 | Same Execution Plan 	 |
+|---------------------------------|-------------|----------------|----------------|----------------|-----------------------|
+| MongoDB Development         	   | ✅         	 | ✅        	     | ✅          	   | ✅        	     | 🔴                	   |
+| MongoDB Production 	            | ✅         	 | ✅           	  | ✅         	    | ✅        	     | 🔴                	   |
```
addaleax, line 75: What's an execution plan? Because, in the example here, they _would_ typically end up using the same queryPlan

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,145 @@
...
+//  +----------------------------------------+
+//            Node(source)
+```
+
+A Node MAY contain parent nodes and children nodes, through specific **components**. A Node that
+doesn't contain any parent node, but contains children nodes is called the **root** node, and
+represents the whole query.
+
+Nodes MAY have additional components that contain metadata for that node. Components MAY have
+references to other Nodes and other components.
```
addaleax, line 119: We should clarify that components are an ordered list in our case, because I'd expect the typical behavior in a component pattern to be that of an unordered list

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,145 @@
...
+
+Nodes with components MAY build a tree like structure, resembling an Abstract Syntax Tree. Nodes MUST
+NOT refer to themselves either directly or through one of it's childrens, avoiding circular references.
```
addaleax, line 122:
```suggestion
NOT refer to themselves either directly or through one of it's children, avoiding circular references.
```

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
kmruiz: Well in this case, in Production we do have an index, but in Development we don't, so they lead to different execution plans (prod = IXSCAN, dev = COLLSCAN) but queries are still equivalent. That's what I was thinking. WDYT?

addaleax: Ok, so the point of this table isn't that they don't have the same execution plan between both queries (they do!), it's that they don't have the same execution plan between both environments?

Then I'd say it's a bit odd how this table works, normally when you have two rows there should either be a possibility of distinct values in each row or the cells for the rows should be merged to clarify that that can't happen

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,147 @@
...
+//  ^                                        ^
+//  +----------------------------------------+
+//            Node(source)
+```
+
+A Node MAY contain parent nodes and children nodes, through specific **components**. A Node that
+doesn't contain any parent node, but contains children nodes is called the **root** node, and
+represents the whole query.
+
+All components in a node MUST be stored in a sorted list. The sorting criteria is left to the specific
```
addaleax, line 118:
```suggestion
All components in a node MUST be stored in a sorted list. The sorting criteria are left to the specific
```

but I'm a bit confused by the terminology, "sorted list" means that the list is actively sorted, "ordered list" just means that the order of elements matters, do we really actively sort components?

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
kmruiz: Ah I get your point now. You are right. I'm thinking how else I can model this. I'll push a suggestion. Do you have a better idea in your mind?

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
addaleax: (In other words: Sorted lists are unordered lists, because there's already a single order, and inserting a new element will always add it at the same position because that position is determined through sorting)

kmruiz: Ah, you are totally right. Probably here we can just summarise as "Components MUST be stored in an ordered list inside a Node.

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
kmruiz: Just tried with a more complex table, but I think it's more clear.

addaleax: It is! :)

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Specification
+
+[BSON](https://bsonspec.org/spec.html) is a binary format that is used to communicate between the
+MongoDB Client (through a driver) and a MongoDB Cluster. MQL BSON (from now on we will just say BSON) 
+is a superset of the original BSON types.
+
+A BSON Type represents the data type inferred from the original source code or from a MongoDB sample
+of documents. A BSON Type MUST be consumable by a MongoDB Cluster and it's serialization MUST be
```
addaleax, line 23:
```suggestion
of documents. A BSON Type MUST be consumable by a MongoDB Cluster and its serialization MUST be
```

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+## META
+
+The keywords "MUST", "MUST NOT", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY"
+and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Specification
+
+[BSON](https://bsonspec.org/spec.html) is a binary format that is used to communicate between the
+MongoDB Client (through a driver) and a MongoDB Cluster. MQL BSON (from now on we will just say BSON) 
+is a superset of the original BSON types.
```
addaleax, line 20: It's a subset? `{ $akjsdhkj: {} }` is valid BSON but not valid MQL

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+A BSON Type represents the data type inferred from the original source code or from a MongoDB sample
+of documents. A BSON Type MUST be consumable by a MongoDB Cluster and it's serialization MUST be
+BSON 1.1 compliant.
+
+### Primitive BSON Types
+
+#### [BsonString](/main/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/BsonType.kt#L58)
+
+A BsonString defines a sequence of characters independently of its locale or encoding. A BsonString MUST be
+encodable to a UTF-8 String.
```
addaleax, line 31: I think what these sentences boil down to "A BonString is a sequence of Unicode characters."

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+MongoDB Client (through a driver) and a MongoDB Cluster. MQL BSON (from now on we will just say BSON) 
+is a superset of the original BSON types.
+
+A BSON Type represents the data type inferred from the original source code or from a MongoDB sample
+of documents. A BSON Type MUST be consumable by a MongoDB Cluster and it's serialization MUST be
+BSON 1.1 compliant.
+
+### Primitive BSON Types
+
+#### [BsonString](/main/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/BsonType.kt#L58)
```
addaleax, line 28: I'd either drop the line numbers of turn these into permalinks, but referring to `main` directly will inevitably result in the specific location becoming outdated

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+
+Represents the absence of a value. 
+
+#### [BsonAny](/main/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/BsonType.kt#L123)
+
+Represents any possible type. Essentially, all type is a subtype of BsonAny.
+
+#### [BsonAnyOf](/main/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/BsonType.kt#L132)
+
+Represents an intersection of types. For example, BsonAnyOf([BsonString, BsonInt32]).
```
addaleax, line 80:
```suggestion
Represents an union of types. For example, BsonAnyOf([BsonString, BsonInt32]). 
```

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
kmruiz: It's a valid BsonObject right, even if it's not a valid query.

kmruiz: I mean, I just repeated what you said, let me rephrase: a BsonType is how we represent parts of an MQL query, but it's not necessarily a valid query per se. For example, a BsonString is not a valid query, but it's still a valid type. This specific document models the types, independently of their use in a query.

kmruiz: When I mean superset I mean because there are semantics that BSON does not have, like BsonAnyOf btw.

addaleax: Alright, I'd maybe mention that as an example here then :)

kmruiz: Done, thanks for pointing this out!

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,151 @@
...
+
+A 128bit floating point number. In Java it's mapped to BigDecimal.
+
+#### BsonNull
+
+Represents the absence of a value. 
+
+#### BsonAny
+
+Represents any possible type. Essentially, all type is a subtype of BsonAny.
```
lerouxb, line 76: nit: `all type is a subtype` ->`all types are subtypes` or `every type is a subtype`

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,150 @@
...
+### Query validation and support
+
+A query MAY be valid for a target MongoDB Cluster if the MongoDB Cluster can consume the query once it
+is translated to a consumable dialect by the target Cluster. However, a query MAY be unsupported by the 
+cluster if it doesn't have the capabilities to fulfill the query request.
+
+For example, let's consider the following query, in pseudocode:
+
+```java
+collection.aggregate(AtlasSearch(text().eq("baby one more time")))
```
lerouxb, line 30: 😆

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,150 @@
...
+We will consider two queries equivalent, independently of the query structure, if the following conditions
+apply:
+
+* They MUST be **valid** by the same set of target clusters.
+* They MUST be **supported** by the same set of target clusters.
+* They MUST return the same subset of results for the same input data set.
+* They MAY be sourced from the same dialect.
+* They MAY lead to equivalent **execution plans** for the same target cluster.
+
+We will consider two execution plans equivalent if the cluster query planner lead to the same list
```
lerouxb, line 54: nit: s/lead/leads

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
kmruiz: images

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,150 @@
...
+| `collection.find({"bookName": myBookName})`   | Production 	                    | 🟢         	 | 🟢           	 | N         	  | mongosh        	     | IXSCAN                	   |
+
+**🟢 They are equivalent because they are valid, supported and return the same result set in all clusters even if the dialect is different.**
+
+
+### Query Nodes
+
+A [MQL Node or a Node for short](/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt) 
+represents a set of semantic properties of a MongoDB query or a subset of it. Nodes MUST NOT be specific to
+a single source dialect, but MAY contain semantics that are relevant for processing.
```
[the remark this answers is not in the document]
himanshusinghs: discussed over call - it captures the case of a partial query and hence "MAY"

packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md
```diff
@@ -0,0 +1,150 @@
...
+//  ^                                        ^
+//  +----------------------------------------+
+//            Node(source)
+```
+
+A Node MAY contain parent nodes and child nodes, through specific **components**. A Node that
+doesn't contain any parent node, but contains child nodes is called the **root** node, and
+represents the whole query.
+
+Components MUST be stored in an ordered list inside a Node. Nodes MAY have additional components that contain metadata for
```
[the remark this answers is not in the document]
himanshusinghs: discussed over call - ordered list is not a sorted list which is what I was mistaking the term with.

packages/mongodb-mql-model/src/docs/md/mql-component/mql-component.md
```diff
@@ -0,0 +1,121 @@
+# MQL Component
+---------------
+
+## Abstract
+
+This specification documents the structure of an MQL Component from a mixed perspective of both
+the original source code and the target server that might run the query. It is primarily aimed
+to provide developers of dialects and linters a common and flexible structure for code processing.
```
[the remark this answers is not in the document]
himanshusinghs: Resolved with recent commit

packages/mongodb-mql-model/src/docs/md/bson-type/bson-type.md
```diff
@@ -0,0 +1,164 @@
...
+case the `baseType` can not be inferred, it MUST be BsonAny.
+
+### Type Assignability
+
+Assignable types MUST not change the semantics of a query when they are swapped. Let's say that
+we have a query $Q$, and two variants, $Q_A$ and $Q_B$, where $Q_A$ and $Q_B$ differ on the specified type
+in either a field or a value reference.
+
+We will say that type $A$ is assignable to type $B$ if $Q_A$ and $Q_B$ are 
+[equivalent queries](/main/packages/mongodb-mql-model/src/docs/md/mql-query/mql-query.md#query-equivalence).
```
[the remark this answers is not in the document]
himanshusinghs: Resolved with recent commit

himanshusinghs (approved):

PR #95, opened 2024-11-21
chore:  Ensure namespace checking is correct for aggregations INTELLIJ-141

kmruiz:
### Checklist


packages/mongodb-linting-engine/src/main/kotlin/com/mongodb/jbplugin/linting/NamespaceCheckingLinter.kt
```diff
@@ -128,7 +128,8 @@ object NamespaceCheckingLinter {
             databaseDoesNotExistParser,
             collectionDoesNotExistParser,
             databaseIsNotKnown,
-            noCollectionSpecified
+            noCollectionSpecified,
+            otherwise { emptyList() }
```
himanshusinghs, line 132: I am curious why this is needed when we already have `.orElse` chained on :136

kmruiz: It's the same issue we had in #93, it will fail the parser if some other run is a success.

himanshusinghs (approved):

PR #96, opened 2024-11-21
feat: adds support for parsing Aggregates.project stage INTELLIJ-126

himanshusinghs:
## Description
Includes:
- Parsing of $project built with Aggregates.project, limited to, as defined in scope:
  - Projections.include
  - Projections.exclude
  - Projections.fields
- Autocomplete support for fields in Projections.include, Projections.exclude and Projections.fields
- Linter inspections for field existence and type checks in Projections

### Checklist

### DIFF
--- a/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
+++ b/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
@@ -58,6 +58,8 @@ data class Node<S>(
 
     inline fun <reified C : Component> components(): List<C> = components.filterIsInstance<C>()
 
+    fun componentsWithChildren(): List<HasChildren<*>> = components.filterIsInstance<HasChildren<*>>()
+
     inline fun <reified C : Component> hasComponent(): Boolean = component<C>() != null
 
     fun withTargetCluster(cluster: HasTargetCluster): Node<S> = copy(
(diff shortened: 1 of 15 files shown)


packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
```diff
@@ -58,6 +58,8 @@ data class Node<S>(
 
     inline fun <reified C : Component> components(): List<C> = components.filterIsInstance<C>()
 
+    fun componentsWithChildren(): List<HasChildren<*>> = components.filterIsInstance<HasChildren<*>>()
```
kmruiz, line 61: I think we can safely change this to:

```suggestion
    fun componentsWithChildren(): List<HasChildren<S>> = components.filterIsInstance<HasChildren<S>>()
```

It will have better type checking. Just a quick note: this is not recursive, so it will only get the first level of children. Is this expected?

kmruiz (approved):

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
himanshusinghs: yea this one is only expected to filter the components that have children. the recursive gathering is taken care of by the parser method.

PR #98, opened 2024-11-22
feat: support for new accumulators INTELLIJ-128

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
+++ b/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
@@ -11,7 +11,6 @@ import com.mongodb.jbplugin.mql.BsonAnyOf
 import com.mongodb.jbplugin.mql.BsonArray
 import com.mongodb.jbplugin.mql.BsonBoolean
 import com.mongodb.jbplugin.mql.BsonInt32
-import com.mongodb.jbplugin.mql.BsonString
 import com.mongodb.jbplugin.mql.BsonType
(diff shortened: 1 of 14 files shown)


packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
```diff
@@ -543,6 +645,44 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
...
     }
 
+    private fun parseComputedExpression(element: PsiElement, createsNewField: Boolean = true): HasValueReference<PsiElement> {
+        val (constant, value) = element.tryToResolveAsConstant()
+        return HasValueReference(
+            when {
+                constant && value is String -> HasValueReference.Computed(
+                    element,
+                    type = ComputedBsonType(
+                        BsonAny,
```
himanshusinghs, line 655:
```suggestion
                        BsonString,
```

himanshusinghs: This I think should be a BsonString because the final expression is written as a String.

himanshusinghs (approved):
Yayy looks perfect - I added a few small suggestions.

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
kmruiz: Yeah but the underlying type of the expression is any possible type right? It's just convenient that MongoDB uses a string for the expression language, but the type is not a string right?

PR #101, opened 2024-11-25
chore: tests and minor suggestions for INTELLIJ-128

himanshusinghs:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
+++ b/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
@@ -734,10 +734,10 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
                         )
                     )
                 )
-                constant && value != null -> HasValueReference.Constant(
+                constant -> HasValueReference.Constant(
                     element,
                     value,
-                    value.javaClass.toBsonType(value)
+                    value?.javaClass.toBsonType(value)
                 )
                 !constant && element is PsiExpression -> HasValueReference.Runtime(
                     element,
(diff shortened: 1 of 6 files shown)


packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
```diff
@@ -612,7 +613,7 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
         val valueExpr = expression.argumentList.expressions.getOrNull(1) ?: return null
 
         val fieldName = keyExpr.tryToResolveAsConstantString()
-        val accumulatorExpr = parseComputedExpression(valueExpr, createsNewField = false)
+        val accumulatorExpr = parseComputedExpression(valueExpr)
```
himanshusinghs, line 616: Realised that for the _id field, the underlying string will always point to a field from schema so I switched createsNewField to false and it turns out that then it is false always, so I removed it entirely.

himanshusinghs: Actually, realised that _id can also be null so that needs to be accounted for. I added another [commit](https://github.com/mongodb/intellij/pull/101/commits/d39254509f882f8242fba6eefc884aad51d8d2ad) to address that.

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
```diff
@@ -715,26 +716,20 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
...
-    private fun parseComputedExpression(element: PsiElement, createsNewField: Boolean = true): HasValueReference<PsiElement> {
+    private fun parseComputedExpression(element: PsiElement): HasValueReference<PsiElement> {
         val (constant, value) = element.tryToResolveAsConstant()
         return HasValueReference(
             when {
                 constant && value is String -> HasValueReference.Computed(
                     element,
                     type = ComputedBsonType(
-                        BsonAny,
+                        BsonString,
```
himanshusinghs, line 726: This is the same suggestion as in the original PR

kmruiz: Ah yes, I remember answering here. The thing is that while how we define the expression is a string, the type of the value is not a string right? For example "$year" is not a string, is an int32, so using a string here is not accurate. I feel that until we can infer the type of the expression, we should just assume any. WDYT?

himanshusinghs: keeping BsonAny here because the actual type is not relevant for now, will revert this particular change

PR #103, opened 2024-11-27
chore(ci): cleanup gha workflows

nirinchev:
## Description

This is a general cleanup/sanitation of gha workflows. It bumps some 3rd party workflows to latest and ensures we only reference them using commit shas (per [github's recommendations](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)). Additionally, it moves our usage of the devtoolsbot user to the github app.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -204,3 +204,5 @@ ij_yaml_keep_line_breaks = true
 ij_yaml_space_before_colon = false
 ij_yaml_spaces_within_braces = true
 ij_yaml_spaces_within_brackets = true
+tab_width = 2
+indent_size = 2
--- a/.github/workflows/draft-release.yaml
+++ b/.github/workflows/draft-release.yaml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
     inputs:
       versionBump:
-        description: 'Version bump'
+        description: "Version bump"
         type: choice
         required: true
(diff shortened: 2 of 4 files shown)


himanshusinghs (approved):

.editorconfig
```diff
@@ -203,4 +203,6 @@ ij_yaml_keep_indents_on_empty_lines = false
 ij_yaml_keep_line_breaks = true
 ij_yaml_space_before_colon = false
 ij_yaml_spaces_within_braces = true
-ij_yaml_spaces_within_brackets = true
\ No newline at end of file
+ij_yaml_spaces_within_brackets = true
+tab_width = 2
```
kmruiz, line 207: Is this change on purpose? It will change the tab size of all our code, not only YAML.

kmruiz (changes requested):

.editorconfig
nirinchev: Hm... that's in the yaml/yml section though - does intellij not respect the extensions in .editorconfig?

kmruiz: Ah wait, you are totally right. Didn't see that part of the diff, was collapsed. All good!

kmruiz (approved):

PR #110, opened 2024-12-10
fix(ci): fix packaging of the segment key INTELLIJ-148

kmruiz:
### Checklist


buildSrc/src/main/kotlin/com.mongodb.intellij.plugin-component.gradle.kts
```diff
@@ -120,12 +120,14 @@ jmhReport {
...
+    if (pluginBundle.enableBundle.get() == true) {
+        register("buildProperties", WriteProperties::class) {
+            group = "build"
 
-        destinationFile.set(project.layout.projectDirectory.file("src/main/resources/build.properties"))
-        property("pluginVersion", rootProject.version)
-        property("segmentApiKey", System.getenv("BUILD_SEGMENT_API_KEY") ?: "<none>")
+            destinationFile.set(project.layout.projectDirectory.file("src/main/resources/build.properties"))
+            property("pluginVersion", rootProject.version)
+            property("segmentApiKey", System.getenv("BUILD_SEGMENT_API_KEY") ?: "<none>")
```
nirinchev, line 129: Should we fail the build here - otherwise if we misconfigure the build environment, we can end up with a bundle without a segment key.

nirinchev (approved):
Minor question, otherwise looks reasonable.

buildSrc/src/main/kotlin/com.mongodb.intellij.plugin-component.gradle.kts
kmruiz: True, that is exactly what happened tbf when we modernised the build configuration with jetbrains best practices.

The thing is that we want to be able to build a version without access to segment. Ideally, we will enable Segment only in "main" and disable segment in the "beta" channel. We need to think more about this.

nirinchev: Perhaps we fail the build if that environment variable is not present and we set it to "<none>" for beta/local builds?

buildSrc/src/main/kotlin/com.mongodb.intellij.plugin-component.gradle.kts
```diff
@@ -120,12 +120,14 @@ jmhReport {
 }
 
 tasks {
-    register("buildProperties", WriteProperties::class) {
-        group = "build"
+    if (pluginBundle.enableBundle.get() == true) {
```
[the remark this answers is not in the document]
kmruiz: It checks if this is the plugin, basically, or a plugin component. For the jetbrains-plugin package this is true, so we bundle the segment api key, if it's not (like the datagrip package) it just discards this task.

PR #113, opened 2025-01-08
fix: fixes autocomplete not working in valid positions because of a Java iterable INTELLIJ-158

himanshusinghs:
## Description
AC was not working in cases where a valid MethodCallExpression was having a Java Iterable as an argument. This PR fixes that by handling the case of Java Iterable and traversing upwards to determine whether a caret position qualifies for AC or not.

### Checklist

### DIFF
--- a/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
+++ b/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
@@ -289,8 +289,19 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
                     HasValueReference(valueReference)
                 ),
             )
-        } else if (method.isVarArgs || method.name == "not") {
-            // Filters.and, Filters.or... are varargs
+        } else if (method.name == "and" || method.name == "or" || method.name == "nor") {
+            return Node(
+                filter,
+                listOf(
+                    Named(Name.from(method.name)),
+                    HasFilter(
+                        filter.getVarArgsOrIterableArgs()
+                            .mapNotNull { resolveBsonBuilderCall(it, FILTERS_FQN) }
+                            .mapNotNull { parseFilterExpression(it) },
+                    ),
+                ),
+            )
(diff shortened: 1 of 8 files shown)


packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
```diff
@@ -1039,20 +1047,26 @@ fun PsiMethodCallExpression.getVarArgsOrIterableArgs(): List<PsiExpression> {
...
 }
 
+fun PsiMethodCallExpression.isJavaIterableCallExpression(
+    method: PsiMethod? = fuzzyResolveMethod()
+): Boolean {
+    val isListOfCall = method?.name == "of" &&
+        method.containingClass?.qualifiedName == JAVA_LIST_FQN
+
+    val isArrayAsListCall = method?.name == "asList" &&
+        method.containingClass?.qualifiedName == JAVA_ARRAYS_FQN
```
kmruiz, line 1057: There are other ways to define iterables in Java, maybe instead of checking the class of the method, we can check the return type is a Collection or an array? Collections.singletonList is also another way.

I don't think we need to support all possible factory methods right now, but if the cost of supporting them is low I would try it.

kmruiz (approved):

packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
himanshusinghs: Discussed this over Slack - Checking the return type of the expression is the sure shot way for inferring wether the expression is a Java Iterable or not but then it would also "accidentally" cover cases for iterables that could potentially be modified such as iterables created with `new ArrayList` interface. For now we decided to rename the interface to `isParseableJavaIterable` to abstract away how we identify the iterable. It also now support `Collections.singletonList` ([71a30ef](https://github.com/mongodb/intellij/pull/113/commits/71a30ef0a7eeb64f812036f5d52dd4adf75f2192))

kmruiz (approved):

PR #117, opened 2025-01-16
chore: refactor aggregation parsing logic into dedicated stage based parsers

himanshusinghs:
This PR splits the current aggregation parsing logic into dedicated stage based parsers.

## Description
This PR splits the current aggregation parsing logic into dedicated stage based parsers. It does not affect the current functionality in any way (also proved by our passing tests 🤞) and will be used as a base for project parsing.

### Checklist

### DIFF
--- a/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
+++ b/packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
@@ -267,7 +267,7 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
                 return null // empty, do nothing
             }
 
-            val fieldReference = resolveFieldNameFromExpression(filter.argumentList.expressions[0])
+            val fieldReference = filter.argumentList.expressions[0].resolveFieldNameFromExpression()
             // if it's only 2 arguments it can be either:
             // - in(field, singleElement) -> valid because of varargs, becomes a single element array
             // - in(field, array) -> valid because of varargs
@@ -343,7 +343,7 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
                 return null
             }
             val fieldExpression = filter.argumentList.expressions[0]
-            val fieldReference = resolveFieldNameFromExpression(fieldExpression)
+            val fieldReference = fieldExpression.resolveFieldNameFromExpression()
             val valueReference = HasValueReference.Inferred(
                 source = fieldExpression,
                 value = true,
@@ -360,7 +360,7 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
             )
         } else if (method.parameters.size == 2) {
             // If it has two parameters, it's field/value.
-            val fieldReference = resolveFieldNameFromExpression(filter.argumentList.expressions[0])
+            val fieldReference = filter.argumentList.expressions[0].resolveFieldNameFromExpression()
             val valueReference = resolveValueFromExpression(filter.argumentList.expressions[1])
 
             return Node(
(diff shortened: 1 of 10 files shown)


packages/mongodb-dialects/java-driver/src/main/kotlin/com/mongodb/jbplugin/dialects/javadriver/glossary/JavaDriverDialectParser.kt
```diff
@@ -818,16 +818,6 @@ object JavaDriverDialectParser : DialectParser<PsiElement> {
...
 
-    private fun resolveFieldNameFromExpression(expression: PsiExpression): HasFieldReference.FieldReference<out Any> {
-        val fieldNameAsString = expression.tryToResolveAsConstantString()
-        val fieldReference =
-            fieldNameAsString?.let {
-                FromSchema(expression, it)
-            } ?: HasFieldReference.Unknown
-
-        return fieldReference
-    }
```
himanshusinghs, line 829: I wanted to use this in Spring as well so I created this as an extension method.

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/AggregationStagesParser.kt
```diff
@@ -0,0 +1,69 @@
...
+        val resolvedStageCalls = newAggregationCallArguments.mapNotNull { stageCallExpression ->
+            stageCallExpression.resolveToMethodCallExpression { _, stageCallMethod ->
+                isStageCall(stageCallMethod)
+            }
+        }
+        return resolvedStageCalls.map { stageCall ->
+            val stageCallMethod = stageCall.fuzzyResolveMethod() ?: return@map Node(
+                source = stageCall,
+                components = emptyList()
+            )
```
himanshusinghs, line 38: now that I am reading this, maybe it is better to return null here because even in the components we will not have a Named node which makes this returned node entirely useless for the rest of the consumers.

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/AggregationStagesParser.kt
```diff
@@ -0,0 +1,69 @@
...
+
+            if (MatchStageParser.isMatchStageCall(stageCallMethod)) {
```
kmruiz, line 40: Can we change this method to be an instance method like `parse`? From what I understand with this design, each stage parser will be an object, so if we put this method in an interface and rename it 'canParse' or something like that, we can use the same interface for all parsers and get consistency and change this if to a for loop over an array of instances.

kmruiz (approved):

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/AggregationStagesParser.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Parser for parsing supported patterns of writing an aggregation pipeline.
+ * Supported patterns for writing aggregation calls are the following:
+ * 1. MongoTemplate.aggregate()
+ * 2. MongoTemplate.aggregateStream()
+ *
+ * The AggregationParser concerns itself only with parsing the aggregation related semantics and
+ * leave the rest as a responsibility for the composing unit.
+ */
+class AggregationStagesParser(private val matchStageParser: MatchStageParser) {
```
kmruiz, line 20: Not now necessarily, but in a future this can be an array or a list of StageParser, and change isStageCall and parseAggregationStages to find the first stage that works for the provided PsiMethod.

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/AggregationStagesParser.kt
kmruiz: Returning null is fine if we filter it out later because we don't want to have nulls in the model if possible. The type system should enforce this, as they accept non-nullable types.

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/AggregationStagesParser.kt
himanshusinghs: ahh interesting composition, I will do the update in the follow up PR that I am about to drop

PR #120, opened 2025-01-17
feat(spring): add support for the unwind aggregation stage INTELLIJ-176

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/UnwindStageParser.kt
+++ b/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/UnwindStageParser.kt
@@ -1,0 +1,40 @@
+package com.mongodb.jbplugin.dialects.springcriteria.aggregationstageparsers
+
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiMethod
+import com.intellij.psi.PsiMethodCallExpression
+import com.mongodb.jbplugin.dialects.javadriver.glossary.tryToResolveAsConstantString
+import com.mongodb.jbplugin.dialects.springcriteria.AGGREGATE_FQN
+import com.mongodb.jbplugin.mql.Component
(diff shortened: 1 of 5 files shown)


packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/UnwindStageParser.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.mongodb.jbplugin.mql.Node
+import com.mongodb.jbplugin.mql.components.HasFieldReference
+import com.mongodb.jbplugin.mql.components.Name
+import com.mongodb.jbplugin.mql.components.Named
+
+class UnwindStageParser : StageParser {
+    override fun isSuitableForFieldAutoComplete(
+        methodCall: PsiMethodCallExpression,
+        method: PsiMethod
+    ) = true
```
himanshusinghs, line 18: I think the method name is a little confusing but the idea of this method `isSuitableForFieldAutoComplete` is for parsers to check and confirm, based on their understanding, if the provided method call is a good target for autocompleter to add field autocomplete.

So it will most likely have the same logic as canParse but might also differ (for example it does differ in project parser). In this case, I think we can safely defer to canParse.

himanshusinghs (commented):
In addition to the parser test, maybe we should extend our current inspection and autocomplete test to also check for cases in unwind.

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/UnwindStageParser.kt
kmruiz: Ah cool, I'll fix it then! I think eventually we will need to add more information to this method then, because not all arguments in a method call expression are suitable for autocomplete. But this is something we can do later.

himanshusinghs (approved):

PR #121, opened 2025-01-20
feat(index): default to queryPlanner and allow users to enable executionStats INTELLIJ-187

kmruiz:
### Checklist


packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/backend/MongoshBackend.kt
```diff
@@ -188,6 +188,19 @@ class MongoshBackend(
...
+     *
+     * This function is <b>unsafe</b>, <b>it does not encode input values, so it's sensitive
+     * to code injection</b>. Only use this for well-known, literal constant values that we have
+     * control of.
+     *
+     * @see emitContextValue for dynamic values provided by the user.
+     */
+    fun emitStringLiteral(value: String): MongoshBackend {
+        return emitAsIs("\"$value\"", encode = false)
+    }
```
himanshusinghs, line 202: Technically we could wrap it in Encode.forJavaScript, or not?

himanshusinghs (approved):
LGTM 🚀

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/backend/MongoshBackend.kt
kmruiz: Tried it, but it encodes the quote chars and makes the string invalid. I guess we could encode only the vale and add the quotes later, but IDK if it makes sense tbf.

PR #126, opened 2025-01-22
feat(query-generation): support for the $project stage INTELLIJ-193

kmruiz:
Did a minor refactor on how $match works, to extract it to it's own file. The rest is just the ground work for $project and the other stages.

### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/codeActions/impl/RunQueryCodeActionBridge.kt
```diff
@@ -124,14 +114,6 @@ internal object RunQueryCodeAction : MongoDbCodeAction {
         )
     }
 
-    private fun shouldShowRunGutterIcon(node: Node<PsiElement>): Boolean {
```
kmruiz, line 127: I removed this method because it was legacy, for when an invalid query would get to here, but now this doesn't happen anymore and we can now generate aggregates.

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/aggr/Aggregate.kt
```diff
@@ -24,33 +21,43 @@ fun <S> Node<S>.isAggregate(): Boolean {
...
+        emitContextValue(resolveValueReference(value, field))
+        emitObjectValueEnd()
+    }
 
     return this
 }
+
+private val NON_DESTRUCTIVE_STAGES = setOf(
+    Name.MATCH
+)
```
himanshusinghs, line 59: should this also include Name.PROJECT?

himanshusinghs (approved):

PR #128, opened 2025-01-23
chore(ci): address codeql issues in gha workflows

nirinchev:
## Description

This addresses most issues reported by codeql. The remaining ones are related to using `mongodb-js/devtools-shared/actions/setup-bot-token@main` instead of pinning to a commit. I feel that this should be treated as a first-party action and be exempt from requiring commit pinning, but am open to be convinced otherwise

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -3,6 +3,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
 
+permissions:
+  contents: read
+  pull-requests: write
+
 jobs:
   changelog:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-pr-title.yml
+++ b/.github/workflows/check-pr-title.yml
@@ -1,7 +1,19 @@
 name: "Check PR Title"
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled, converted_to_draft, edited]
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
+      - unlabeled
+      - converted_to_draft
+      - edited
+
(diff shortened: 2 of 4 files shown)


.github/workflows/check-changelog.yml
```diff
@@ -3,6 +3,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
 
+permissions:
+  contents: read
+  pull-requests: write
```
kmruiz, line 8: Why are we adding additional permissions to this workflow? We don't need them for checking the changelog and blocking a PR right?

.github/workflows/check-pr-title.yml
```diff
@@ -1,24 +1,36 @@
...
+      - reopened
+      - ready_for_review
+      - labeled
+      - unlabeled
+      - converted_to_draft
+      - edited
+
+permissions:
+  contents: read
+  pull-requests: read
```
kmruiz, line 16: Same thing here, we don't want workflows to have additional permissions unless they are needed.

kmruiz (changes requested):

.github/workflows/check-changelog.yml
nirinchev: Maybe I'm misunderstanding something, but https://github.com/mongodb/intellij/blob/d7678ff7524e3c1d9ccd1be0cae1f47efad57ab3/.github/workflows/check-changelog.yml#L14 requires write permissions for PRs to add the labels, doesn't it? Also, not that this is not adding additional permissions as the default is `write` for everything - instead we're explicitly specifying only the components we actually need and defaulting to `none` for everything else.

kmruiz: Ah, then you are right. Now it makes sense that with the default permissions it works. Got it!

kmruiz (approved):

PR #130, opened 2025-01-23
feat(query-generation): support for the unwind stages INTELLIJ-195

kmruiz:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ MongoDB plugin for IntelliJ IDEA.
 ## [Unreleased]
 
 ### Added
+* [INTELLIJ-195](https://jira.mongodb.org/browse/INTELLIJ-195) Add support for $unwind.
 * [INTELLIJ-194](https://jira.mongodb.org/browse/INTELLIJ-194) Add support for $addFields.
 * [INTELLIJ-193](https://jira.mongodb.org/browse/INTELLIJ-193) Add support for generating aggregates with $match and $project.
 * [INTELLIJ-189](https://jira.mongodb.org/browse/INTELLIJ-189) Add support for generating update queries.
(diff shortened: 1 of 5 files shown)


packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/query/References.kt
```diff
@@ -29,9 +29,9 @@ fun <S> MongoshBackend.resolveValueReference(
 
 fun <S> MongoshBackend.resolveFieldReference(fieldRef: HasFieldReference<S>) =
     when (val ref = fieldRef.reference) {
-        is FromSchema -> registerConstant(ref.fieldName)
-        is Inferred<S> -> registerConstant(ref.fieldName)
-        is HasFieldReference.Computed<S> -> registerConstant(ref.fieldName)
+        is FromSchema -> registerConstant(ref.displayName)
```
kmruiz, line 32: Changed to use displayName because fieldName removes any prefixed '$' in the field, and in unwind it's mandatory. Because displayName holds the information "as the user wrote it" makes sense to use it here. We were using fieldName because this part was done before displayName was introduced.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ MongoDB plugin for IntelliJ IDEA.
 ## [Unreleased]
 
 ### Added
+* [INTELLIJ-195](https://jira.mongodb.org/browse/INTELLIJ-195) Add support for $unwind.
```
himanshusinghs, line 8: Sorry I did not notice this before but I guess we wanna be a little more precise about what support we added for this and the other item?

himanshusinghs (approved):
Yay 🚀

CHANGELOG.md
kmruiz: sounds good, let me just make another commit making this more precise

PR #131, opened 2025-01-23
feat(query-generation): add support for $sort INTELLIJ-196

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/query/References.kt
+++ b/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/query/References.kt
@@ -1,5 +1,6 @@
 package com.mongodb.jbplugin.dialects.mongosh.query
 
+import com.mongodb.jbplugin.dialects.mongosh.backend.ContextValue
 import com.mongodb.jbplugin.dialects.mongosh.backend.MongoshBackend
 import com.mongodb.jbplugin.mql.BsonAny
 import com.mongodb.jbplugin.mql.BsonString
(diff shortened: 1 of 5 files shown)


packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/query/References.kt
```diff
@@ -13,18 +14,28 @@ import com.mongodb.jbplugin.mql.components.HasValueReference
...
-    )
+): ContextValue {
+    return when (val ref = valueRef.reference) {
+        is HasValueReference.Computed -> {
+            val node = ref.type.expression
+            val fieldRef =
+                node.component<HasFieldReference<S>>()
+                    ?: return registerVariable("queryField", ref.type.baseType)
+
+            resolveFieldReference(fieldRef)
```
kmruiz, line 25: Added this because I realised that we weren't supporting Computed values. This is something that we will need to improve whenever we support Computed values that are more than just a field reference.

himanshusinghs: This is going to be interesting because now we can finally see if the root bsonType will be helpful in generating these queries or not. Right now we only take one HasFieldReference component but there could be more :D

himanshusinghs: Should we leave a todo with a ticket or were you planning to tackle this as part of group?

himanshusinghs (approved):

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/query/References.kt
kmruiz: AFAIK for now we are only going to support field names as references, no arbitrary expressions, so I think this is kinda of enough for our use cases. I was thinking that we could rework this method in the future whenever we support more types of expressions. I can create a ticket for this, I think it's good to track it.

PR #134, opened 2025-01-30
feat: support generating shell syntax for $group INTELLIJ-197

himanshusinghs:
## Description
This PR adds support for generating shell syntax for $group stage and the supported accumulators.

### Checklist

### DIFF
--- a/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/aggr/Aggregate.kt
+++ b/packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/aggr/Aggregate.kt
@@ -36,6 +36,7 @@ fun <S> MongoshBackend.emitAggregateBody(node: Node<S>, queryContext: QueryConte
             Name.ADD_FIELDS -> emitAddFieldsStage(stage)
             Name.UNWIND -> emitUnwindStage(stage)
             Name.SORT -> emitSortStage(stage)
+            Name.GROUP -> emitGroupStage(stage)
             else -> {}
         }
         emitObjectValueEnd(long = true)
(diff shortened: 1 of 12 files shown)


packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/aggr/Aggregate.kt
```diff
@@ -49,7 +50,12 @@ internal fun <S> MongoshBackend.emitAsFieldValueDocument(nodes: List<Node<S>>, i
         val field = node.component<HasFieldReference<S>>() ?: continue
         val value = node.component<HasValueReference<S>>() ?: continue
 
-        emitObjectKey(resolveFieldReference(field))
+        emitObjectKey(
+            resolveFieldReference(
+                fieldRef = field,
+                fieldUsedAsValue = false,
+            )
+        )
```
himanshusinghs, line 58: @kmruiz I adopted this approach to make MongoshBackend understand when to prefix the field name with a `$` and when not to. The idea is if a field reference is used in the value place then it is supposed to be true and the emitted field will be prefixed with `$` otherwise not.

packages/mongodb-dialects/mongosh/src/main/kotlin/com/mongodb/jbplugin/dialects/mongosh/aggr/Aggregate.kt
```diff
@@ -63,6 +69,7 @@ private val NON_DESTRUCTIVE_STAGES = setOf(
     Name.ADD_FIELDS,
     Name.UNWIND,
     Name.SORT,
+    Name.GROUP,
```
himanshusinghs, line 72: When I was writing this I was not sure if `$group` is a non-destructive stage. But I assume this is a non-destructive stage as it does not generally perform a complex accumulation or a cross collection lookup. Let me know if this was not the idea for this collection `NON_DESTRUCTIVE_STAGES`. :D

kmruiz: It's perfect, I was considering destructing stages all of which do write into mongodb like $out or $merge.

himanshusinghs: Ok that makes sense, thank you for confirming :D

kmruiz (approved):

PR #135, opened 2025-02-03
feat: adds support for parsing `Sort` objects attached to `Query` objects in spring data mongodb INTELLIJ-199

himanshusinghs:
## Description
This PR adds support for parsing `Sort` objects attached to `Query` objects in spring data mongodb using `.with` call.

We have re-used the entire `Sort` object parsing logic from `SortStageParser` which is why there is a little refactor over there (without any logic changes) and on top of that built a `QuerySortParser` to handle the specific cases of writing a query expression.

Test case coverage includes:
- parsing test
- field inspection test
- autocomplete test
- query generation test which was already present

### Checklist

### DIFF
--- a/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
+++ b/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
@@ -173,7 +173,7 @@ class SortStageParser : StageParser {
                             ?: return@flatMapIndexed emptyList<Node<PsiElement>>()
 
                     return@flatMapIndexed parseSortObjectArgument(
-                        sortCreationMethodCall,
+                        sortCreationMethodCallPassedToAnd,
                         forcedDirection,
                         reverseCount
                     )
--- a/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
+++ b/packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
@@ -90,208 +90,211 @@ class SortStageParser : StageParser {
         }
     }
 
-    private fun parseDirectionArgumentAndFieldsToSortCriteria(
-        directionArgumentExpression: PsiExpression,
-        fieldArgumentExpressions: List<PsiExpression>,
-        directionForcedFromParentChain: Name?,
-        reverseCountForcedFromParentChain: Int
-    ): List<Node<PsiElement>> {
-        val directionName = directionArgumentExpression.resolveToDirectionName().derivedDirection(
-            directionForcedFromParentChain,
-            reverseCountForcedFromParentChain
-        )
-        return fieldArgumentExpressions.map { fieldExpression ->
-            createSortCriteria(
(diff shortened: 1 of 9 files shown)


packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
```diff
@@ -90,208 +90,211 @@
...
+                else if (method.name == "reverse" && forcedDirection == null) {
+                    reverseCount += 1
+                    return@flatMapIndexed emptyList<Node<PsiElement>>()
+                }
+                // This is how we chain additional Sort criteria chains onto the current chain.
+                // This chained Sort criteria will inherit the forcedDirection and reverseCount from
+                // the current chain
+                else if (method.name == "and") {
+                    val firstArgumentExpression = methodCall.argumentList.expressions.getOrNull(0)
+                    val sortCreationMethodCallPassedToAnd =
```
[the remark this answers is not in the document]
kmruiz: This seems to be true. I see that this is the result of calling a function. If we don't use the variable maybe we can make explicit that we require to resolve to a sortCreationCall but we don't care about the value yet?

```suggestion
                    if (firstArgumentExpression?.resolveToSortCreationCall() == null) return@flatMApIndexed emptyList<Node<PsiElement>>()
```

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/SpringCriteriaDialectParserTest.kt
```diff
@@ -72,6 +72,84 @@ class Repository {
         }
     }
 
+    @ParsingTest(
+        fileName = "Book.java",
+        """
+import org.springframework.data.domain.Sort;import org.springframework.data.mongodb.core.MongoTemplate;
```
kmruiz, line 78: Probably we will need to do a cleanup here, in the same way we did on the other project, so we don't need to apply the imports. And likely we can consolidate to have the assertion DSL in all projects.

Nothing urgent, I'll add it to the backlog.

kmruiz (approved):

packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/aggregationstageparsers/SortStageParser.kt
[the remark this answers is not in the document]
himanshusinghs: Ok so it turns out I actually passed the wrong variable during refactor. That is also what got caught in the test cases. I will pass the correct one and push.

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/SpringCriteriaDialectParserTest.kt
himanshusinghs: I will do some cleanup already so that we don't accumulate this stuff. Thank you for pointing this out :D

himanshusinghs: Ahh I see, the other cleanup was in JB plugin. Then it makes sense that we do it separately.

PR #140, opened 2025-02-06
feat(spring): support for Reactive Spring Data INTELLIJ-230

kmruiz:
Depends on #138 

We already had some support as the syntax is really close to the ordinary sync Spring Data. However, we added tests to ensure we officially support, at least, all the functionalities with support for sync and registered the interfaces into our parser so they are taken into consideration.

### Checklist


packages/mongodb-dialects/spring-criteria/src/main/kotlin/com/mongodb/jbplugin/dialects/springcriteria/SpringCriteriaDialectParser.kt
```diff
@@ -696,10 +698,12 @@ private val INTERFACES_WITH_QUERY_METHODS = arrayOf(
     "ExecutableMapReduceOperation",
     "ExecutableRemoveOperation",
     "ExecutableUpdateOperation",
+    "ReactiveMongoOperations",
```
himanshusinghs, line 701:
```suggestion
```
Mentioned already above. :)

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/reactive/ReactiveAggregationStagesParserTest.kt
```diff
@@ -0,0 +1,212 @@
...
+record Book() {}
+
+class Repository {
+    private final ReactiveMongoTemplate template;
+    
+    public Repository(ReactiveMongoTemplate template) {
+        this.template = template;
+    }
+    
+    public AggregationResults<Book> allReleasedBooks() {
```
himanshusinghs, line 85:
```suggestion
    public Flow.Publisher<Book> allReleasedBooks() {
```
how important is the return type here? I think that the parsers can still parse correctly without having to specify correct return type still I wonder.

himanshusinghs (approved):
Amazing work 🚀

packages/mongodb-dialects/spring-criteria/src/test/kotlin/com/mongodb/jbplugin/dialects/springcriteria/reactive/ReactiveAggregationStagesParserTest.kt
kmruiz: You are correct that for our parser is not very relevant but wanted to change the code to make it as correct as possible, so if at some point becomes relevant we don't need to fix 100 tests.

himanshusinghs: updated the rest as well

PR #142, opened 2025-02-07
chore: add ESR role to each named operation INTELLIJ-209

kmruiz:
### Checklist


packages/mongodb-mql-model/src/test/kotlin/com/mongodb/jbplugin/mql/components/NamedTest.kt
```diff
@@ -0,0 +1,319 @@
+package com.mongodb.jbplugin.mql.components
+
+import com.mongodb.jbplugin.mql.BsonAny
+import com.mongodb.jbplugin.mql.BsonType
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class NamedTest {
+
+    @Test
```
kmruiz, line 10: I'm not using `@ParameterizedTest` here because kotlin-test won't support that.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/codeActions/impl/runQuery/RunQueryModal.kt
```diff
@@ -31,7 +31,6 @@ import com.mongodb.jbplugin.mql.toNonNullableType
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.runBlocking
 import org.bson.types.ObjectId
-import org.jetbrains.letsPlot.commons.intern.filterNotNullValues
```
kmruiz, line 34: This API is deprecated in 2025 EAP, so getting rid of this.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/Named.kt
```diff
@@ -5,78 +5,218 @@
...
+    },
+    BOTTOM_N("bottomN") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
+    MAX("max") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
+    MIN("min") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
```
himanshusinghs, line 186: Not sure about these - these are group operators so they should not have a query role I guess, technically IRRELEVANT?

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/Named.kt
```diff
@@ -5,78 +5,218 @@
...
+    SORT("sort") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
+    ASCENDING("ascending") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
+    DESCENDING("descending") {
+        override fun queryRole(bsonType: BsonType): QueryRole = QueryRole.SORT
+    },
```
himanshusinghs, line 210: I suppose we consider these are SORT because one can chain .sort on queries as well which is when these will become relevant.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/Named.kt
kmruiz: After some investigation you are right. I _really_ thought they could use an index, but no. Changed them to irrelevant.

PR #150, opened 2025-02-12
chore: adds data distribution based on sample docs in CollectionSchema INTELLIJ-214

himanshusinghs:
### Checklist

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -22,5 +22,5 @@ jobs:
       - name: Check Changelog
         uses: dangoslen/changelog-enforcer@204e7d3ef26579f4cd0fd759c57032656fdf23c7 # 3.6.1
         with:
-          skipLabels: "no release notes"
+          skipLabels: no release notes
           missingUpdateErrorMessage: Please add an entry in CHANGELOG.md or apply the 'no release notes' label to skip this check.
--- a/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
+++ b/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
@@ -1,7 +1,5 @@
 package com.mongodb.jbplugin.mql
 
-import org.bson.Document
-
 data object JsonArray
 data object JsonObject
 data object JsonUndefined
@@ -40,17 +38,6 @@ data class DataDistribution(private val distribution: Map<Field, Map<Value, Occu
                     val fieldPath = if (parentPath.isEmpty()) field else "$parentPath.$field"
                     val fieldDistribution = distribution.getOrPut(fieldPath) { mutableMapOf() }
                     when (value) {
-                        is Document -> {
-                            fieldDistribution[JsonObject] =
-                                fieldDistribution.getOrDefault(JsonObject, 0) + 1
-                            val mappedValue = value.toMap()
-                            populateDistributionForList(
-                                listOf(mappedValue),
-                                distribution,
-                                fieldPath
-                            )
-                        }
-
                         is Map<*, *> -> {
                             fieldDistribution[JsonObject] =
                                 fieldDistribution.getOrDefault(JsonObject, 0) + 1
@@ -109,11 +96,6 @@ data class DataDistribution(private val distribution: Map<Field, Map<Value, Occu
                 paths.add(fieldPath)
 
                 when (value) {
-                    is Document -> {
-                        val mappedValue = value.toMap()
-                        getAllFieldPathsFromDocument(mappedValue, paths, fieldPath)
-                    }
-
                     is Map<*, *> -> {
                         val mappedValue = value.map { it.key.toString() to it.value }.toMap()
                         getAllFieldPathsFromDocument(mappedValue, paths, fieldPath)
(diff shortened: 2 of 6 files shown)


.github/workflows/check-changelog.yml
```diff
@@ -22,5 +22,5 @@ jobs:
       - name: Check Changelog
         uses: dangoslen/changelog-enforcer@204e7d3ef26579f4cd0fd759c57032656fdf23c7 # 3.6.1
         with:
-          skipLabels: "no release notes"
+          skipLabels: no release notes
```
himanshusinghs, line 25: This was necessary otherwise the action was expecting label applied with double quotes.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
```diff
@@ -0,0 +1,156 @@
+package com.mongodb.jbplugin.mql
+
+import org.bson.Document
```
kmruiz, line 3: The MQL model shouldn't depend on org.bson because it's not multiplatform. Can't we use an ordinary map instead?

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
```diff
@@ -0,0 +1,156 @@
...
+
+data object JsonArray
+data object JsonObject
```
kmruiz, line 6: Is there a way to use the MQL BsonType instead?

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
```diff
@@ -0,0 +1,156 @@
...
+            sampleDocs: List<Map<Field, Value>>,
+            distribution: MutableMap<Field, MutableMap<Value, OccurrenceCount>>,
+            parentPath: Field = "",
+        ) {
+            for (sampleDoc in sampleDocs) {
+                for ((field, value) in sampleDoc) {
+                    val fieldPath = if (parentPath.isEmpty()) field else "$parentPath.$field"
+                    val fieldDistribution = distribution.getOrPut(fieldPath) { mutableMapOf() }
+                    when (value) {
+                        is Document -> {
```
kmruiz, line 43: We are using this check twice but I don't see where we are constructing a BSON Document. Are we expecting to get BSON Documents from the outside? A query will not return BSON Documents because we got rid of it a few weeks ago.

kmruiz (changes requested):
Great work!

Only left a few comments that are blocking but should be straightforward to fix.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: We definitely can, as it is just a placeholder. I will make the change.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: The comment above is related to this so I am answering here. I noticed that we expect a `Document` when we are constructing the schema out of sample of documents which is why I handled it here. But as you said, we are not mapping to a Document, I will quickly verify and remove this check if that is the case.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: Just realised that we use these when creating keys for the distribution for objects and they keys are generally String, which is why I won't be able to use a BsonType here. I also don't want to use static strings such as "object" or "array" as they might come up as field in a document.

kmruiz:
Discussed everything on a call. I'll approve the PR so you can merge once you do the changes we discussed.

Great work!

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: Talked about this over Slack. We will be keeping them with a note that they are not meant to be used outside as they are just placeholder values.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: Removed the usage of this

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/DataDistribution.kt
himanshusinghs: Removed the usage of Document and also updated GetCollectionSchema to remove the usage there.

PR #152, opened 2025-02-13
chore(ci): ensure spec changes when the mql model or engines change INTELLIJ-228

kmruiz:
## Description

A new GitHub check that verifies if, for the mql-model or the mql-engine, we changed anything and we didn't update the spec itself. Essentially, it verifies that, if a kotlin file in the project has been changed and there is a src/docs folder, you need to change the docs folder accordingly.
 
These checks can be disabled by adding the `skip-spec-change` label in the PR. This is useful if the change is small and does not require changes in the spec (like a bug fix).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ MongoDB plugin for IntelliJ IDEA.
 ## [Unreleased]
 
 ### Added
+* [INTELLIJ-232](https://jira.mongodb.org/browse/INTELLIJ-219) Consider selectivity of a field over cardinality when suggesting indexes.
 * [INTELLIJ-219](https://jira.mongodb.org/browse/INTELLIJ-219) Implement index consolidation of a query and it's siblings.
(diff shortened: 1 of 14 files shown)


.github/workflows/quality-check.yaml
```diff
@@ -13,6 +13,45 @@ permissions:
...
+          else
+            echo "SKIP=false" >> $GITHUB_ENV
+          fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run Spec Update Check
+        if: github.ref != 'refs/heads/main' && env.SKIP != 'true'
+        run: |
+          ./gradlew --quiet --console=plain checkSpecUpdates
```
himanshusinghs, line 53: I think this needs to happen in a separate workflow, just like "Check Changelog" otherwise when we label / unlabel, that event won't trigger this job.

himanshusinghs (changes requested):

himanshusinghs (approved):

PR #154, opened 2025-02-14
chore: get rid of bson in the access adapter INTELLIJ-207

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
```diff
@@ -187,22 +187,22 @@ class CachedQueryService(
                 val readModel by query.source.project.service<DataGripBasedReadModelProvider>()
                 val buildInfo = readModel.slice(dataSource, BuildInfo.Slice)
 
-                queryWithDb.withTargetCluster(
```
kmruiz, line 190: At some point we bugged this by not storing the modified query (data classes are immutable).

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
```diff
@@ -62,19 +62,20 @@ fun waitFor(timeout: Duration, interval: Duration, condition: () -> Boolean) {
...
 ) {
     var attempt = 1
     waitFor(timeout, Duration.ofMillis(50)) {
         val result = runCatching {
             fn(attempt++)
-            true
-        }.getOrDefault(false)
+        }
+
+        result.exceptionOrNull()?.printStackTrace(System.err)
```
kmruiz, line 73: In case there is an exception, this will print it, so we get more information about what is happening.

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/MongoDbDriver.kt
```diff
@@ -27,6 +25,10 @@ sealed interface QueryResult<S> {
     data class Run<S>(val result: S) : QueryResult<S>
 }
 
+class ConnectionString(hostInfo: List<String>) {
+    val hosts = hostInfo.map { it.replace("mongodb://", "").replace("mongodb+srv://", "") }
```
kmruiz, line 29: The java driver removes the prefix, so we are doing this here too in case we need it. This ensures that the slices work for any driver even if they don't remove the prefix.

himanshusinghs: I assume this will work just fine for standalone implementations as well and is server version independent?

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/HasRunCommand.kt
```diff
@@ -0,0 +1,9 @@
+package com.mongodb.jbplugin.mql.components
+
+import com.mongodb.jbplugin.mql.Component
+
+data class HasRunCommand<S>(
```
himanshusinghs, line 5: more of a question that is not blocking at all:
I don't see a runCommand command being written in the user code. Assuming that is true, can we get rid of `S` here or put it definitely typed as Unit, as there will be no source?

himanshusinghs (approved):
Yayy - no bson

packages/mongodb-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/MongoDbDriver.kt
kmruiz: Yes! it should at least.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/components/HasRunCommand.kt
kmruiz: That's a good point. I don't think we are going to support runCommand in the short term, but I don't know if maintaining the generic source is that expensive to get rid of it and then adding it whenever we want to implement runCommand. WDYT?

PR #169, opened 2025-03-20
chore: major refactor to linters to support our new view model INTELLIJ-247

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
```diff
@@ -0,0 +1,132 @@
...
+                    val dataSource = fileInExpression?.dataSource ?: return@runReadAction
+                    fileInExpression.dialect ?: return@runReadAction
+
+                    val inspectionsViewModel by expression.project.service<InspectionsViewModel>()
+                    val queryService by expression.project.service<CachedQueryService>()
+                    val query = queryService.queryAt(expression) ?: return@runReadAction
+
+                    if (fileInExpression.virtualFile != null && dataSource.isConnected()) {
+                        val settings = buildSettings(query)
+                        val problems = IntelliJBasedQueryInsightsHolder(coroutineScope, holder, inspectionsViewModel, ::afterInsight)
```
himanshusinghs, line 82:
```suggestion
                        val problems = IntelliJBasedQueryInsightsHolder<I>(coroutineScope, holder, inspectionsViewModel, ::afterInsight)
```
I think the `I` is not automatically inferred here because there is no parameter passed  to the constructor that is bound to `I`.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.intellij.openapi.components.Service
+import com.intellij.psi.PsiElement
+import com.mongodb.jbplugin.linting.QueryInsight
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.withContext
+
+@Service(Service.Level.PROJECT)
+class InspectionsViewModel {
+    val insights: MutableStateFlow<List<QueryInsight<PsiElement, *>>> = MutableStateFlow(emptyList())
```
himanshusinghs, line 12:
```suggestion
    private val mutableInsights: MutableStateFlow<List<QueryInsight<PsiElement, *>>> = MutableStateFlow(emptyList())
    val insights: StateFlow<List<QueryInsight<PsiElement, *>>>
      get() = mutableInsights
```
nit: because the insights won't be modified from outside, maybe we can expose the immutable flow?

himanshusinghs (approved):
Great refactor Kevin also super helpful changes on the test utils. 🚀

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
kmruiz: I think in this case it is automatically inferred because ::afterInsight is a method that uses I, so actual implementations of the class have to provide the I.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
kmruiz: Changed 😄

PR #170, opened 2025-03-25
chore: restore the previously selected data source INTELLIJ-273

himanshusinghs:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/implementations/PersistentConnectionPreferences.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/implementations/PersistentConnectionPreferences.kt
@@ -7,32 +7,32 @@ package com.mongodb.jbplugin.editor.services.implementations
 
 import com.intellij.openapi.components.*
 import com.intellij.openapi.project.Project
-import com.mongodb.jbplugin.editor.services.ToolbarSettings
+import com.mongodb.jbplugin.editor.services.ConnectionPreferences
 import java.io.Serializable
 
 // Setting this service as a PROJECT service ensures that our state is saved per Project
 @Service(Service.Level.PROJECT)
 @State(
-    name = "com.mongodb.jbplugin.editor.ToolbarSettings",
+    name = "com.mongodb.jbplugin.editor.ConnectionPreferences",
     storages = [
         Storage(
             // File used to persist these settings per project
-            value = "MongodbToolbarSettings.xml",
+            value = "MongodbConnectionPreferences.xml",
             // Setting roamingType to DEFAULT ensures that these settings are carried forward during IntelliJ updates
             // as well
             roamingType = RoamingType.DEFAULT
         )
     ]
 )
-internal class ToolbarSettingsStateComponent :
-    SimplePersistentStateComponent<PersistentToolbarSettings>(PersistentToolbarSettings())
+internal class ConnectionPreferencesStateComponent :
+    SimplePersistentStateComponent<PersistentConnectionPreferences>(PersistentConnectionPreferences())
 
-internal class PersistentToolbarSettings :
+internal class PersistentConnectionPreferences :
     BaseState(),
     Serializable,
-    ToolbarSettings {
+    ConnectionPreferences {
     override var dataSourceId by string(null)
-    override var database by string(ToolbarSettings.UNINITIALIZED_DATABASE)
+    override var database by string(ConnectionPreferences.UNINITIALIZED_DATABASE)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


gradle/libs.versions.toml
```diff
@@ -9,7 +9,7 @@ java-target="21"
 versions-plugin = "0.51.0"
 jacoco = "0.8.12"
 test-retry = "1.6.0"
-intellij-plugin = "2.2.1"
+intellij-plugin = "2.4.0"
```
kmruiz, line 12: Did it improve the performance of building the plugin?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
```diff
@@ -87,16 +89,23 @@ class ConnectionStateViewModel(
     private val coroutineScope: CoroutineScope
 ) : DataSourceManager.Listener, JdbcDriverManager.Listener {
     internal var connectionSaga = ConnectionSaga(project, ::emitSelectedConnectionChanged)
-    val connectionState = MutableStateFlow(ConnectionState.default())
+    internal val mutableConnectionState = MutableStateFlow(ConnectionState.default())
+    val connectionState: StateFlow<ConnectionState> = mutableConnectionState
```
kmruiz, line 93:
```suggestion
    val connectionState = mutableConnectionState.asStateFlow()
```

It's a bit shorter and we don't need to specify the type twice.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/implementations/PersistentConnectionPreferences.kt
```diff
@@ -7,39 +7,39 @@ package com.mongodb.jbplugin.editor.services.implementations
...
 // Setting this service as a PROJECT service ensures that our state is saved per Project
 @Service(Service.Level.PROJECT)
 @State(
-    name = "com.mongodb.jbplugin.editor.ToolbarSettings",
+    name = "com.mongodb.jbplugin.editor.ConnectionPreferences",
     storages = [
         Storage(
             // File used to persist these settings per project
-            value = "MongodbToolbarSettings.xml",
+            value = "MongodbConnectionPreferences.xml",
```
himanshusinghs, line 20: A side effect of this change is that with the next update of the plugin, the saved settings from the Toolbar won't get propagated to the new settings. We can leave the underlying file as is so that the old saved settings get propagated but for consistency I anyways did the change. We can also implement a migration mechanism but given that our plugin is not even public just yet I would recommend skipping that as well and accepting that we will start a clean slate on the next plugin update for the internal users. Let me know what you think @kmruiz?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
himanshusinghs: Addressed this in my [recent commit](https://github.com/mongodb/intellij/pull/170/commits/cf48496cdfb1688172a935bf7141d80b44ce7efc).

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
```diff
@@ -87,16 +89,23 @@ class ConnectionStateViewModel(
...
 
     init {
         coroutineScope.launch {
             val dataSources = project.getDataSourceService().listMongoDbDataSources()
-            connectionState.emit(ConnectionState(dataSources, Empty))
+            mutableConnectionState.emit(ConnectionState(dataSources, Empty))
+
+            val toolbarSettings = project.getToolbarSettings()
```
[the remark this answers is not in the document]
himanshusinghs: Addressed this in the [recent commit](https://github.com/mongodb/intellij/pull/170/commits/cf48496cdfb1688172a935bf7141d80b44ce7efc). Maybe just read the commit itself and not the entire set of changes to make it easier to go through it :)

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/services/implementations/PersistentConnectionPreferences.kt
kmruiz: I agree with your approach overall, let's just rename it. I think it's reasonable to force the users to connect once again manually.

kmruiz (approved):

PR #171, opened 2025-03-25
chore: show performance warnings in the sidebar INTELLIJ-275

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
@@ -1,0 +1,69 @@
+package com.mongodb.jbplugin.ui.viewModel
+
+import com.intellij.openapi.application.EDT
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.fileEditor.FileEditorManager
+import com.intellij.openapi.fileEditor.FileEditorManagerEvent
+import com.intellij.openapi.fileEditor.FileEditorManagerListener
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.VirtualFile
+import com.mongodb.jbplugin.editor.services.MdbPluginDisposable
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+data class EditorState(
+    val focusedFiles: List<VirtualFile>,
+    val openFiles: List<VirtualFile>
+) {
+    companion object {
+        fun default(): EditorState {
+            return EditorState(emptyList(), emptyList())
+        }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun fileClosed(source: FileEditorManager, file: VirtualFile) {
+        rebuildEditorState(source)
+    }
+
+    private fun rebuildEditorState(manager: FileEditorManager) {
+        val openFiles = manager.openFiles.toList().distinctBy { it.canonicalPath }
+        val focusedFiles = manager.selectedEditors.mapNotNull { it.file }.distinctBy { it.canonicalPath }.toList()
+
+        mutableEditorState.tryEmit(EditorState(focusedFiles, openFiles))
```
himanshusinghs, line 57: Ok if IIUC, the tryEmit might fail silently when there is a buffer overflow. I feel like the fileOpened selectionChanged and fileClosed events are crucial such that if we happen to skip any we might end up with an inconsistent SidePanel state. Take for example - a selection changed event but we failed to emit the event and the SidePanel still displays the warnings from old file.

Maybe for this usecase its better to just use plain emit?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
```diff
@@ -21,6 +28,27 @@ class InspectionsViewModel {
...
+    suspend fun visitQueryOfInsightInEditor(insight: QueryInsight<PsiElement, *>) {
+        withContext(Dispatchers.EDT) {
+            val vFile = insight.query.source.containingFile.virtualFile
+
+            val manager = FileEditorManager.getInstance(insight.query.source.project)
+            val editorOfFile = manager.selectedTextEditor?.takeIf { it.virtualFile == vFile }
+                ?: manager.openTextEditor(OpenFileDescriptor(insight.query.source.project, vFile, 0), true)
+                ?: return@withContext
+
+            editorOfFile.caretModel.moveToOffset(insight.query.source.textOffset)
```
himanshusinghs, line 48: Perhaps we should consider offloading this to CodeEditorViewModel? That would keep the concerns together and leave us with a re-usable logic, should we ever find ourselves using it again.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
kmruiz: Hmmm, that's a good point. I think it would be unlikely to block because the logic we have in collect is pretty straightforward, but I get your point. I think we can run a coroutine in IO so we don't block the EDT thread.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
kmruiz: Yes, that's a great idea! Working on it.

PR #174, opened 2025-04-01
chore: change the query inlay to show index status INTELLIJ-271

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
```diff
@@ -32,16 +32,25 @@ import kotlinx.coroutines.withContext
...
     protected abstract fun buildSettings(query: Node<PsiElement>): Settings
     protected abstract fun emitFinishedInspectionTelemetryEvent(problemsHolder: ProblemsHolder)
     protected open fun afterInsight(queryInsight: QueryInsight<PsiElement, I>) {}
 
+    override fun inspectionStarted(
+        session: LocalInspectionToolSession,
+        isOnTheFly: Boolean
+    ) {
+        val inspectionViewModel by session.file.project.service<InspectionsViewModel>()
+        inspectionViewModel.startInspectionSessionOf(session.file, inspection)
```
kmruiz, line 51: This will clean up old inspections. We need this because we didn't have a way to remove solved inspections.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -57,6 +61,21 @@ class CodeEditorViewModel(
...
 
+    suspend fun reanalyzeRelevantEditors() {
+        withContext(Dispatchers.IO) {
+            withinReadAction {
+                val allPsiFiles = editorState.value.focusedFiles.mapNotNull {
+                    it.findPsiFile(project)
+                }
+
+                for (psiFile in allPsiFiles) {
+                    InlayHintsPassFactoryInternal.forceHintsUpdateOnNextPass()
```
kmruiz, line 72: While this is called Internal, it's not an Internal class. It's necessary so also Inlay Hints are reprocessed. If we don't call this, we only run a pass for the inspections.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
```diff
@@ -284,6 +289,52 @@ class ConnectionStateViewModel(
...
+        val query = context.query
+
+        if (dataSource == null || project == null || query == null) {
+            null
+        } else {
+            Triple(dataSource, project, query)
+        }
+    }
+
+    override fun requestFinished(context: Context) {
```
kmruiz, line 310: This fixes an issue with us not detecting when the cluster changes. This is registered as a listener to a topic that is triggered when a script finished successfully. We do here some wishful thinking, and check if the script uses some know mutation commands (createIndex, dropIndex, insert...) and then:

1. We increase the dataSource modification count. This refreshes the ReadModelProvider cache.
2. We reanalyse the current scope, so the inspections run again with the new state of the cluster.

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
```diff
@@ -110,6 +117,26 @@ data class Node<S>(
     override fun toString(): String {
         return "Node(components=$components)"
     }
+
+    fun queryHash(): Int {
```
kmruiz, line 121: It seems that two nodes might not be equivalent depending on the "pass" during analysing. This was affecting the Inlay, as it couldn't use the cached results of the index verification from other run processes like the inspections. This simple query hash is used in the ExplainQuery slice to have a consistent hash based on the referenced fields in a query.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
```diff
@@ -0,0 +1,138 @@
...
+        }.getOrDefault(NotRun)
+
+        val (icon, text, tooltip) = when (explainPlan) {
+            CollectionScan ->
+                Triple(
+                    Icons.indexWarningIcon,
+                    InspectionsAndInlaysMessages.message("inlay.indexing.coll-scan.text"),
+                    InspectionsAndInlaysMessages.message("inlay.indexing.coll-scan.tooltip")
+                )
+            is IndexScan ->
```
himanshusinghs, line 107: probably ddon't need `is` here and below. All static objects.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
himanshusinghs: Good catch - totally forgot about this.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/AnalysisScopeViewModel.kt
```diff
@@ -26,6 +26,12 @@ class AnalysisScopeViewModel(
         }
     }
 
+    suspend fun reanalyzeCurrentScope() {
+        mutableAnalysisScope.tryEmit(mutableAnalysisScope.value)
```
himanshusinghs, line 30: This seems odd - why do we emit the current value itself, again?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
himanshusinghs: I guess running the inspections does not trigger inlay hint updates

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -57,6 +61,21 @@ class CodeEditorViewModel(
...
+                    DaemonCodeAnalyzer.getInstance(project).restart(psiFile)
```
himanshusinghs, line 73: What about this - does this also not trigger the inlay hint updates? I always thought of this as hard reboot 😄

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
```diff
@@ -284,6 +289,52 @@ class ConnectionStateViewModel(
...
+        val (dataSource, project, query) = readContext(context) ?: return
+
+        if (likelyChangesDataSource(query)) {
+            dataSource.incModificationCount()
+
+            val codeEditorViewModel by project.service<AnalysisScopeViewModel>()
+            runBlocking {
+                codeEditorViewModel.reanalyzeCurrentScope()
```
himanshusinghs, line 318:
```suggestion
                codeEditorViewModel.reanalyzeRelevantEditors()
```
I did not find anything such as reanalyzeCurrentScope. Maybe a mistype?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
himanshusinghs: Yea I recall this from our conversation. Thought we did not need it because we introduced the Refresh icon button. Did not consider that inlay hints will have to refreshed on changes. Great stuff here <3

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
```diff
@@ -18,6 +20,14 @@ class InspectionsViewModel {
     private val mutableOpenCategories = MutableStateFlow<InspectionCategory?>(null)
     val openCategories = mutableOpenCategories.asStateFlow()
 
+    fun startInspectionSessionOf(psiFile: PsiFile, inspection: Inspection) {
+        val allOtherInspections = insights.value.filter {
+            !(it.inspection == inspection && it.query.source.containingFile.isEquivalentTo(psiFile))
+        }
+
+        mutableInsights.tryEmit(allOtherInspections)
```
himanshusinghs, line 28: I feel like this should also be using the emit inside a suspended coroutine as it looks like a crucial update which when missed will lead to odd state in the UI. What do you think?

himanshusinghs (approved):
Solid work Kevin! I marked a few things to consider. Let me know what you think about them.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
himanshusinghs: Please ignore - I scrolled down to the class implementation and saw that you changed it from object to data class.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/AnalysisScopeViewModel.kt
kmruiz: This is because for anyone listening to events in the flow, it should retrigger their logic again. I felt that this way was cleaner, as we inverse the dependency: adding more consumers doesn't require changing this function.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/ConnectionStateViewModel.kt
kmruiz: Ah yes, the typo is actually in the variable name, the class is correct. Fixing it in a new commit. Likely missed the variable name while I was refactoring.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/InspectionsViewModel.kt
kmruiz: That sounds good, we can block until the event is emitted.

PR #175, opened 2025-04-01
feat: added DatabaseComboBox for spring related dialect INTELLIJ-258

himanshusinghs:
## User facing changes
- Adds the DatabaseComboBox which will always be visible in the SidePanel
- The value selected in the DatabaseComboBox is taken into consideration only when the inferred dialect is Spring related dialect (changes in MongoDBVirtualDataSourceProvider)

## Preview

https://github.com/user-attachments/assets/14666d11-3d12-4a50-846b-4786d557117c

### Checklist


kmruiz:
My concern with this approach is that for projects not using Spring, the database selector is not meaningful and might be misleading. Can we at least mention that this is a default database, used only when we can't infer what database the user is querying in code?

PR #180, opened 2025-04-04
chore: migrates namespace related insights to SidePanel INTELLIJ-274

himanshusinghs:
## Description
Extracts namespace related inspections earlier built into NamespaceCheckingLinter and creates individual inspection implementations:
- DatabaseDoesNotExistInspection
- NoDatabaseInferredInspection
- CollectionDoesNotExistInspection
- NoCollectionSpecifiedInspection

Each of these inspections implements a primary action of "Choose Connection" which essentially opens up the SidePanel and expands the connection combobox.

## Previews:
|Insights from Database and Collection existence check inspections|image|
|---|---|
|Insights for Database not detected|image|
|---|---|
|Insights for Collection not specified|image|
|---|---|
|Primary action|[Video](https://github.com/user-attachments/assets/32ec33f8-1aaa-49ef-becb-fc0a4833854d)|

### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/components/utilities/ControlledComboBox.kt
```diff
@@ -0,0 +1,269 @@
...
+    // this gets called - like it was in the setPopupExpanded fun below.
+    // onPopupStateChange: (Boolean) -> Unit = {},
+    popupContent: @Composable () -> Unit,
+) {
+    var popupExpanded by remember { mutableStateOf(false) }
+    var chevronHovered by remember { mutableStateOf(false) }
+
+    LaunchedEffect(comboBoxExpanded) {
+        popupExpanded = comboBoxExpanded
+    }
```
himanshusinghs, line 100: This is what I was thinking would be controversial because generally we would never do that in React but I needed to do this because the listeners in the rendering that works with popupExpanded (the state coming from parent) never updates themselves even when the state changes. An example [is here](https://github.com/mongodb/intellij/pull/180/files#diff-6bee01706ad9279b2a296618d783194829f55096305102f9b26a37e0ba46b747R143-R149), this particular snippet always works on an outdated value if the state but with a local copy of state this seems to work fine. I don't understand Compose enough to dig out what was causing this but when time permits, I will get to the bottom of it.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/components/connection/ConnectionBootstrapCard.kt
```diff
@@ -315,3 +343,51 @@ internal fun useConnectionStatus(connection: LocalDataSource): State<ConnectionS
...
+    return LocalConnectionComboBoxState.current ?: ConnectionComboBoxState.default()
+}
+
+@Composable
+internal fun useSidePanelEventSubscription(
+    event: SidePanelEvents,
+    onEvent: (SidePanelEvents) -> Unit,
+) {
+    val viewModel by useViewModel<SidePanelViewModel>()
+    viewModel.subscribeToSidePanelEvents { emittedEvent ->
```
kmruiz, line 388: This is a side-effect that will be called each time the parent composable is called. We might want to run it in a LaunchedEffect with a dependency on event, so it's only called once per event subscription.

himanshusinghs: Ahh that is true - totally missed this.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/insightAction/ChooseConnectionInsightAction.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.mongodb.jbplugin.linting.QueryInsight
+import com.mongodb.jbplugin.meta.service
+import com.mongodb.jbplugin.ui.viewModel.SidePanelViewModel
+
+class ChooseConnectionInsightAction : InsightAction {
+    override val displayName = "Choose a connection"
+
+    override suspend fun apply(insight: QueryInsight<PsiElement, *>) {
+        val viewModel by insight.query.source.project.service<SidePanelViewModel>()
+        viewModel.openSidePanel()
```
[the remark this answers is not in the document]
himanshusinghs: Makes sense. I will do a small fixup on this.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/components/connection/ConnectionBootstrapCard.kt
himanshusinghs: Applied a fix for this here [0a5c69e](https://github.com/mongodb/intellij/pull/180/commits/0a5c69edaea4ab2ad3496b1e1b3c2c190bfe9e00)

PR #181, opened 2025-04-04
chore: support for a new scope: AllInsights INTELLIJ-281

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/analysisScope/AnalysisScope.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/analysisScope/AnalysisScope.kt
@@ -1,35 +1,83 @@
 package com.mongodb.jbplugin.inspections.analysisScope
 
+import com.intellij.openapi.fileTypes.FileTypeManager
+import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.openapi.vfs.findPsiFile
 import com.intellij.psi.PsiElement
+import com.intellij.psi.search.FileTypeIndex
+import com.intellij.psi.search.GlobalSearchScope
+import com.mongodb.jbplugin.dialects.Dialect
+import com.mongodb.jbplugin.dialects.javadriver.glossary.JavaDriverDialect
+import com.mongodb.jbplugin.dialects.springcriteria.SpringCriteriaDialect
+import com.mongodb.jbplugin.dialects.springquery.SpringAtQueryDialect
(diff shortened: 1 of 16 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/analysisScope/AnalysisScope.kt
```diff
@@ -1,35 +1,83 @@
...
+            return allFiles
+                .mapNotNull { it.findPsiFile(project) }
+                .map { file -> file to allDialects.firstOrNull { it.isUsableForSource(file) } }
+                .filter { it.second != null }
+                .map {
+                    it.first.virtualFile.apply {
+                        putUserData(Keys.attachedDialect, it.second!!)
+                    }
+                }
+        }
```
himanshusinghs, line 81: This is small nit. I the sideeffect of attaching the dialect onto the relevant files gets hidden under the name `getAdditionalFilesInScope`, perhaps it should be more explicit. I was thinking something like:
```
val files = getAdditionalFilesInScope()
ensureAttachedDialect(files)
```
Also this op might be better suited for CodeEditorViewModel or maybe something else to centralise it because I think right now we are having similar operations being done in different places.

But feel free to add this to a task which we can tackle later if you think the idea is reasoable.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/AnalysisScopeViewModel.kt
```diff
@@ -1,30 +1,145 @@
...
         coroutineScope.launch {
             val codeEditorViewModel by project.service<CodeEditorViewModel>()
-            codeEditorViewModel.editorState.value.run(::refreshAnalysisScopeIfNecessary)
-            codeEditorViewModel.editorState.collectLatest(::refreshAnalysisScopeIfNecessary)
+            codeEditorViewModel.editorState.value.run { refreshAnalysisScopeIfNecessary() }
+            codeEditorViewModel.editorState.collectLatest { refreshAnalysisScopeIfNecessary() }
+        }
+    }
+
+    suspend fun changeScope(scope: AnalysisScope) = withContext(Dispatchers.IO) {
```
himanshusinghs, line 78: I think we have to take care of this fun potentially called multiple times simultaneously which essentially means that the status updates that we are emitting needs to check if the status is for the scope which is currently set right now and only then update it otherwise discard it. Similar to how we do in ConnectionStateViewModel.

himanshusinghs (changes requested):
Amazing work Kevin! I left one small nit and one point which might consideration.

himanshusinghs (approved):

PR #182, opened 2025-04-07
chore: support for current query scope and fix side panel icon INTELLIJ-283

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
```diff
@@ -128,8 +132,18 @@ private object QueriesInFileCollector : InlayHintsCollector {
...
+            inlayFactory.withTooltip(tooltip, it)
+        }.let {
+            inlayFactory.withCursorOnHover(it, Cursor(Cursor.HAND_CURSOR))
+        }.let {
+            inlayFactory.onClick(it, Left) { _, _ ->
+                val analysisScopeViewModel by element.project.service<AnalysisScopeViewModel>()
+                runBlocking {
+                    element.project.mongoDbSidePanel.show()
+                    analysisScopeViewModel.changeScopeToCurrentQuery()
+                }
```
himanshusinghs, line 145: I had a similar situation on my PR as well where when I was doing runBlocking it was occasionally freezing up the UI. It was weird because the action is not that much time taking but maybe we should consider runBlocking on some other thread and switching the context for UI interactions?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/components/utilities/Card.kt
```diff
@@ -46,3 +48,8 @@ fun Card(category: CardCategory, title: String, body: @Composable () -> Unit) {
         }
     }
 }
+
+object Card {
+    val backgroundColor = if (JBColor.isBright()) JBColor.GRAY.toComposeColor() else Color(0x39, 0x3B, 0x40)
+    val secondaryBackgroundColor = if (JBColor.isBright()) JBColor.DARK_GRAY.toComposeColor() else Color(0x2B, 0x2D, 0x30)
+}
```
himanshusinghs, line 55: <3

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -75,6 +87,15 @@ class CodeEditorViewModel(
         rebuildEditorState(source)
     }
 
+    override fun caretPositionChanged(event: CaretEvent) {
+        val allCarets = event.editor.caretModel.allCarets.map {
+            CaretView(it.editor.virtualFile, it.offset)
+        }
+
+        val toEmit = mutableEditorState.value.copy(carets = allCarets)
+        mutableEditorState.tryEmit(toEmit)
```
himanshusinghs, line 96: Do you think we should be wrapping this up in coroutineScope.launch? Seems like an important emit to be lost when overbufferred.

himanshusinghs (approved):
Looks perfect 🚀

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
kmruiz: That's a really good idea. I'll do the change, just in case and see if it works.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
kmruiz: I was thinking that this might not be that important because if we overflow, it's likely the user is moving the caret a lot (navigating to somewhere line by line or char by char), so if we lose the intermediate states I think it's reasonable.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
kmruiz: Btw, I will wait for #180 and remove the call to the side panel that we have here in favour of the ViewModel you are introducing.

himanshusinghs (approved):

PR #186, opened 2025-04-15
chore: removes deadcode for old toolbar and related components INTELLIJ-274

himanshusinghs:
also introduces the following changes:
- the Gutter icon won't show up when there is no connected data source in the Sidepanel
- filtering insights were throwing errors occasionally which are now being caught
- CodeEditorViewModel tracks the editor related events and attaches the dialect to the editors

### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
```diff
@@ -68,7 +69,11 @@ class CachedQueryService(
             return decorateWithMetadata(dataSource, attachment.getUserData(queryCacheKey)!!.value)
         }
 
-        val toolbar = attachment.project.getToolbarModel()
+        // TODO: Confirm if instead of ToolbarModel, now the dependency should be the
+        // ConnectionStateViewModel or not. Tested with the applied dependency as well and saw no
+        // effect
+
+        // val toolbar = attachment.project.getToolbarModel()
```
kmruiz, line 76: This is important because dependencies should implement ModificationTracker. The toolbar was implementing it.

Maybe we don't need this anymore because the view models are smarter and know when to reanalyse.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/analysisScope/AnalysisScope.kt
```diff
@@ -7,6 +7,7 @@ import com.mongodb.jbplugin.linting.QueryInsight
 import com.mongodb.jbplugin.meta.service
 import com.mongodb.jbplugin.meta.withinReadActionBlocking
 import com.mongodb.jbplugin.ui.viewModel.CodeEditorViewModel
+import fleet.util.logging.logger
```
kmruiz, line 10: Maybe this is not the import we are looking for.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -57,24 +65,47 @@ private val allDialects: List<Dialect<PsiElement, Project>> = listOf(
...
 
         rebuildEditorState(manager)
         messageBusConnection.subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, this)
+        messageBusConnection.subscribe(PsiModificationTracker.TOPIC, this)
         EditorFactory.getInstance().eventMulticaster.addCaretListener(this, disposable)
     }
 
+    override fun modificationCountChanged() {
+        val focusedFile = editorState.value.focusedFile
+        if (focusedFile != null) {
```
kmruiz, line 103: Why are we removing the dialect on changes?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -27,7 +32,9 @@ import com.mongodb.jbplugin.editor.MongoDbVirtualFileDataSourceProvider.Keys
 import com.mongodb.jbplugin.editor.services.MdbPluginDisposable
 import com.mongodb.jbplugin.meta.service
 import com.mongodb.jbplugin.meta.withinReadAction
+import com.mongodb.jbplugin.meta.withinReadActionBlocking
 import com.mongodb.jbplugin.mql.Node
+import fleet.util.logging.logger
```
kmruiz, line 37: Same, this is not the logger we usually use.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
```diff
@@ -88,12 +119,19 @@ class CodeEditorViewModel(
...
+                CaretView(it.editor.virtualFile, it.offset)
+            }
+            val toEmit = mutableEditorState.value.copy(carets = allCarets)
+            mutableEditorState.tryEmit(toEmit)
+        } catch (e: Exception) {
+            // There are a few edge cases where tracking the caret position bubbles an error in
+            // the IDE for null virtualFile, so we catch that here and log it.
+            // Unfortunately it appears that we cannot log the exception as well because the
+            // exception message also tries to access the virtualFile and fails again.
+            log.info("Unable to track caretPositionChanged event")
```
kmruiz, line 133: Use the log message builder when writing logs, it makes them parseable (uses structured logging) and adds additional metadata like the plugin version.

kmruiz (approved):
It feels amazing to remove all of this code!

Left a few comments, feel free to merge when resolved.

Brilliant work!

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
himanshusinghs: There is chance that the modification in the editor removed the import statements in which case we need to update the dialect so we remove what is cached and recheck it.

I suppose your concern is that this might end up happening too many times? I am thinking maybe we can debounce the actual call?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/analysisScope/AnalysisScope.kt
himanshusinghs: Eagle eyes 😄 Thanks for spotting. I will fix this right away.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
himanshusinghs: Ok so because I was not using the right import, the IDE was not complaining about me using the plain string text. Thanks for pointing this out. 👍

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
himanshusinghs: So that is what I am confused about - I implemented the modification tracker in ConnectionStateViewModel which is the replacement for ToolbarModel and it does not make much difference whether I add it or not. Do you recall what was the reason we added the toolbar model as the dependency here?

kmruiz: Ah yes, I do remember. The thing is that for Spring, when you changed the database from the selector, we had to reparse the queries as they were related to another database, so we used the modification tracker to invalidate the cache. In this case, it should behave the same: whether we change the connection or the database, we should invalidate the cache.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/CodeEditorViewModel.kt
himanshusinghs: Perfect - I will keep a close watch on this then.

PR #191, opened 2025-04-24
chore: fix side panel freezing on handling unattached queries INTELLIJ-298

kmruiz:
### Checklist


gradle/libs.versions.toml
```diff
@@ -9,7 +9,7 @@ java-target="21"
 versions-plugin = "0.51.0"
 jacoco = "0.8.12"
 test-retry = "1.6.0"
-intellij-plugin = "2.2.1"
+intellij-plugin = "2.5.0"
```
himanshusinghs, line 12: This might be why the tests are failing. :D

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/codeActions/impl/runQuery/NamespaceSelector.kt
```diff
@@ -133,7 +134,7 @@ class NamespaceSelector(
                 databaseComboBox.isEnabled = true
             }
             is Event.DatabaseSelected -> {
-                coroutineScope.launchChildBackground {
+                coroutineScope.launch {
```
himanshusinghs, line 137:
```suggestion
                coroutineScope.launch(Dispatchers.IO) {
```

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
```diff
@@ -50,7 +50,7 @@ abstract class AbstractMongoDbInspectionBridgeV2<Settings, I : Inspection>(
         isOnTheFly: Boolean
     ) {
         val inspectionViewModel by session.file.project.service<InspectionsViewModel>()
-        runBlocking {
+        coroutineScope.launch {
```
himanshusinghs, line 53: Should we consider doing most of our things (non-UI related) in IO dispatcher? Just so we don't risk clogging the main?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/meta/NodeEx.kt
```diff
@@ -0,0 +1,13 @@
+package com.mongodb.jbplugin.meta
+
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiFile
+import com.intellij.psi.util.findTopmostParentOfType
+import com.mongodb.jbplugin.mql.Node
+
+val Node<PsiElement>.containingFileOrNull: PsiFile?
+    get() = source.findTopmostParentOfType()
```
himanshusinghs, line 10: To ask kevin - when there is no type specified, will it default to getting the file itself?

himanshusinghs (approved):
One small change to pull in and the rest looks perfect 🚀

gradle/libs.versions.toml
himanshusinghs: I noticed the same when I tried updating the plugin to 2.5.0. We might wanna hold off the update until later?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridgeV2.kt
kmruiz: Yeah I think so, it should make everything more reliable. I don't think the cost of dispatching a coroutine is noticeable enough in a UI.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/meta/NodeEx.kt
kmruiz: To answer Himanshu: in this case, Kotlin is smart enough to infer that the type is PsiFile because if you look at the simplified version of the signature of findTopmostParentOfType:

```
fun <T> findTopmostParentOfType(): T?
val Node<PsiElement>.containingFileOrNull: PsiFile = findTopmostParentOfType()
// then the only way it compiles is if it's expanded to:
val Node<PsiElement>.containingFileOrNull: PsiFile = findTopmostParentOfType<PsiFile>()
```

gradle/libs.versions.toml
kmruiz: I think I managed to reproduce the issues with the tests, I will try to fix them. It's a QoL fix, but I think it's worth in the long term because with this plugin version, everything is just slow and we are wasting a lot of time waiting.

PR #192, opened 2025-04-25
fix: fixes inspection hierarchy for inspection runs INTELLIJ-293

himanshusinghs:
- Database inspection will suppress Collection, Field, Field Type and Query inspections
- Collection inspection will suppress Field, Field Type and Query Inspections
- Field inspection will suppress Field Type inspections

Additionally we will display the Index inlay only when Namespace is available in the cluster.

### Checklist

### DIFF
--- a/packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/environmentmismatch/CollectionDoesNotExistInspection.kt
+++ b/packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/environmentmismatch/CollectionDoesNotExistInspection.kt
@@ -6,11 +6,26 @@ import com.mongodb.jbplugin.linting.Inspection.CollectionDoesNotExist
 import com.mongodb.jbplugin.linting.QueryInsight
 import com.mongodb.jbplugin.linting.QueryInsightsHolder
 import com.mongodb.jbplugin.linting.QueryInspection
+import com.mongodb.jbplugin.mql.Namespace
 import com.mongodb.jbplugin.mql.Node
 import com.mongodb.jbplugin.mql.adt.Either
 import com.mongodb.jbplugin.mql.parser.components.knownCollection
 import com.mongodb.jbplugin.mql.parser.filter
 import com.mongodb.jbplugin.mql.parser.parse
+
+fun <D>Namespace.isCollectionAvailableInCluster(
+    dataSource: D,
+    readModelProvider: MongoDbReadModelProvider<D>
+): Boolean {
+    val availableCollections = runCatching {
+        readModelProvider.slice(
+            dataSource,
+            ListCollections.Slice(database)
+        ).collections.map { it.name }
+    }.getOrDefault(emptyList())
+
+    return availableCollections.contains(collection)
+}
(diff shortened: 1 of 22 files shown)


packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/environmentmismatch/CollectionDoesNotExistInspection.kt
```diff
@@ -6,12 +6,27 @@ import com.mongodb.jbplugin.linting.Inspection.CollectionDoesNotExist
...
 import com.mongodb.jbplugin.mql.adt.Either
 import com.mongodb.jbplugin.mql.parser.components.knownCollection
 import com.mongodb.jbplugin.mql.parser.filter
 import com.mongodb.jbplugin.mql.parser.parse
 
+fun <D>Namespace.isCollectionAvailableInCluster(
+    dataSource: D,
+    readModelProvider: MongoDbReadModelProvider<D>
+): Boolean {
+    val availableCollections = runCatching {
```
kmruiz, line 20: nit: this is O(n2) because we are iterating fully twice the collection (for the map and the contains). We can either convert it to a sequence with asSequence, so it's lazy, or use .any { it.name == collection }.

packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/environmentmismatch/DatabaseDoesNotExistInspection.kt
```diff
@@ -6,12 +6,25 @@ import com.mongodb.jbplugin.linting.Inspection.DatabaseDoesNotExist
...
 import com.mongodb.jbplugin.mql.adt.Either
 import com.mongodb.jbplugin.mql.parser.components.knownCollection
 import com.mongodb.jbplugin.mql.parser.filter
 import com.mongodb.jbplugin.mql.parser.parse
 
+fun <D>Namespace.isDatabaseAvailableInCluster(
+    dataSource: D,
+    readModelProvider: MongoDbReadModelProvider<D>,
+): Boolean {
+    val availableDatabases = runCatching {
```
kmruiz, line 20: nit: Same as: https://github.com/mongodb/intellij/pull/192/files#r2060590415

packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/environmentmismatch/CollectionDoesNotExistInspection.kt
himanshusinghs: Probably not O(n2), maybe O(n+n) because the loop is not nested?
But we can definitely club these two instructions together. I will make the change.

kmruiz: True it's O(2n), mistyped :(

kmruiz (approved):

PR #203, opened 2025-04-29
fix: disable actions on queries with unsupported features INTELLIJ-301

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
+++ b/packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
@@ -157,6 +157,10 @@ data class Node<S>(
             }.all { it }
         }
     }
+
+    fun isSupportedBlocking(): Boolean {
+        return runBlocking { isSupported() }
+    }
 }
 
 /**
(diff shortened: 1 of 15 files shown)


packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
```diff
@@ -137,6 +140,26 @@ data class Node<S>(
 
         return allFieldNames.hashCode()
     }
+
+    private val supportedCommands = (CommandType.entries.filter { it.isSupported }).toSet()
+    private val supportedOperations = (Name.entries.filter { it.isSupported }).toSet()
+    suspend fun isSupported(): Boolean {
+        return withContext(Dispatchers.IO) {
+            val isCommandSupportedAsync = async { whenCommandIsAnyOf<S>(supportedCommands).parse(this@Node) }
```
himanshusinghs, line 148: What is this syntax? `this@Node`

himanshusinghs (commented):
Should we add some parsing tests as well? (totally optional though)

packages/mongodb-mql-model/src/main/kotlin/com/mongodb/jbplugin/mql/Node.kt
kmruiz: this@Node evaluates to the `this` object in the closer parent scope of type Node. So basically, it returns this `this` Node, so it can be parsed.

himanshusinghs (approved):

PR #205, opened 2025-05-02
fix(completion): use async completion so it doesn't freeze the IDE

kmruiz:
## Description

It also replaces the usage of deprecated CoroutineScope APIs to use the newer and safer APIs, dispatching the logic in Dispatchers.IO to avoid blocking the UI thread.

### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
@@ -40,9 +40,12 @@ import com.mongodb.jbplugin.settings.pluginSetting
 import com.mongodb.jbplugin.ui.viewModel.AnalysisScopeViewModel
 import com.mongodb.jbplugin.ui.viewModel.InspectionsViewModel
 import com.mongodb.jbplugin.ui.viewModel.SidePanelViewModel
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeoutOrNull
 import org.jetbrains.jewel.bridge.JewelComposePanel
 import java.awt.Cursor
 import javax.swing.JComponent
+import kotlin.time.Duration.Companion.seconds
 
 class MongoDbQueryIndexStatusInlay : InlayHintsProvider<Unit> {
     override val name: String = InspectionsAndInlaysMessages.message("inlay.indexing.name")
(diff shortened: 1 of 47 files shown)


packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
```diff
@@ -86,6 +88,53 @@ fun eventually(
...
+ * @param timeout
+ * @param fn
+ * @param recovery
+ */
+private fun eventuallySuspendable(
+    timeout: Duration = Duration.ofSeconds(10),
+    coroutineScope: TestScope? = null,
+    recovery: () -> Unit = {},
+    fn: suspend (Int) -> Unit,
+) {
```
kmruiz, line 115: We need this because using runBlocking on eventually blocks the entire test so it never finishes. By making this suspendable, the function can yield to other coroutines and works as expected.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inlays/MongoDbQueryIndexStatusInlay.kt
```diff
@@ -99,25 +101,33 @@ private object QueriesInFileCollector : InlayHintsCollector {
         val queryContext = QueryContext.empty(automaticallyRun = true)
 
         val collectionReference = query.component<HasCollectionReference<PsiElement>>()?.reference
-        val skipInlayDecoration = collectionReference !is HasCollectionReference.Known ||
-            !collectionReference.namespace.isValid ||
-            !collectionReference.namespace.isNamespaceAvailableInCluster(
-                dataSource = dataSource,
-                readModelProvider = readModelProvider,
-            )
+        val skipInlayDecoration = runBlocking {
```
himanshusinghs, line 104: The same question for this as well.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/meta/CoroutinesEx.kt
```diff
@@ -55,3 +57,13 @@ fun <T> withinReadActionBlocking(cb: () -> T): T {
         cb()
     }
 }
+
+suspend fun <T> withDefaultTimeoutOrNull(cb: suspend () -> T): T? {
+    return withTimeoutOrNull(2.seconds) {
+        cb()
+    }
+}
```
himanshusinghs, line 65: I am not sure but we probably should think about if its needed to expose this as configurable property as well. I understand that its technical but if we expose it as a property we can avoid having to release a new version just to update the timeout if we see it as too short or too long :)

packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/fixtures/AssertionExtensions.kt
```diff
@@ -62,6 +63,7 @@ fun waitFor(timeout: Duration, interval: Duration, condition: () -> Boolean) {
  * @param fn
  * @param recovery
  */
+@Deprecated("Prefer using TestScope.eventually as it works with coroutines.")
```
himanshusinghs, line 66: We can probably remove this now? Please ignore if there are still usage and the plan is to replace them later.

packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/DataGripBasedReadModelProvider.kt
```diff
@@ -56,27 +60,39 @@ class DataGripBasedReadModelProvider(
...
-    override fun <T : Any> slice(
+    override suspend fun <T : Any> slice(
         dataSource: LocalDataSource,
         slice: Slice<T>,
     ): T {
-        return cachedValues.compute("${dataSource.uniqueId}/${slice.id}") { key, data ->
-            val modificationStamp = data?.first ?: -1
-            val cachedValue = data?.second
+        return withContext(Dispatchers.IO) {
+            withTimeout(10.seconds) {
```
himanshusinghs, line 71: Maybe I am recalling incorrectly but I thought we implemented `maxTimeMs` but I cannot find that in the code at-least. Do we plan to use this as the maxTimeMs on the "client side"? If so should we think of exposing this as a configurable prop?

himanshusinghs (approved):
Amazing work 🚀 Left some questions but they are not blocker.

PR #211, opened 2025-05-08
chore: removes runBlocking usage

himanshusinghs:
### Checklist


packages/mongodb-access-adapter/datagrip-access-adapter/src/main/kotlin/com/mongodb/jbplugin/accessadapter/datagrip/DataGripBasedReadModelProvider.kt
```diff
@@ -66,6 +69,7 @@ class DataGripBasedReadModelProvider(
     override suspend fun <T : Any> slice(
         dataSource: LocalDataSource,
         slice: Slice<T>,
+        onCacheRecalculation: (suspend (T) -> Unit)?,
```
kmruiz, line 72: This seems to be breaking some of our tests because Mockito kind of complains that we are not using a matcher when mocking this, which is fair. I believe it can be easily fixed because it's just providing a `any()` to these functions and in several places is encapsulated in a single method (like QueryInspectionTestUtils).

kmruiz (approved):

PR #218, opened 2025-05-28
chore: additional tests to verify inspections for .iterator and .cursor calls INTELLIJ-317

himanshusinghs:
## Description
Queries terminating with `.iterator` and `.cursor` calls have always been parsed and inspected for insights. The additional test cases are only to confirm the existing behaviour.

The PR adds test cases for:
1. Missing fields insight
2. Field Value type mismatch insight

and each for MongoCursor returned by a `find()` and `aggregate()` call.

### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/inspections/correctness/JavaDriverMongoDbFieldDoesNotExistTest.kt
+++ b/packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/inspections/correctness/JavaDriverMongoDbFieldDoesNotExistTest.kt
@@ -27,6 +27,38 @@ public FindIterable<Document> exampleFind() {
             .getCollection("myCollection")
             .find(eq(<warning descr="Field \"nonExistingField\" does not seem to exist in collection.">"nonExistingField"</warning>, "123"));
 }
+// Additional tests to verify INTELLIJ-317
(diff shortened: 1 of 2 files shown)


packages/jetbrains-plugin/src/test/kotlin/com/mongodb/jbplugin/inspections/correctness/JavaDriverMongoDbFieldDoesNotExistTest.kt
```diff
@@ -27,6 +27,38 @@ public FindIterable<Document> exampleFind() {
             .getCollection("myCollection")
             .find(eq(<warning descr="Field \"nonExistingField\" does not seem to exist in collection.">"nonExistingField"</warning>, "123"));
 }
+// Additional tests to verify INTELLIJ-317
+public MongoCursor<Document> exampleFind1() {
```
nirinchev, line 31: No strong feelings here and happy to keep it as-is if it follows established naming patterns, but would it make sense to name these methods a bit more descriptively? E.g. `findIterator_nonExistentField_showsWarning` - that way a reader doesn't have to read the entire test case source (which is admittedly not very long) to understand what is being tested, under what conditions, and what the expected outcome is.

nirinchev (approved):
Looks reasonable (based on my limited kotlin experience). I left a minor naming convention nit, but that's not blocking.

PR #224, opened 2025-06-04
chore: return a valid mongodb config when disconnected during the query INTELLIJ-332

kmruiz:
### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
+++ b/packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
@@ -9,7 +9,7 @@ import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiManager
 import com.intellij.psi.util.CachedValue
-import com.intellij.psi.util.CachedValueProvider
+import com.intellij.psi.util.CachedValueProvider.Result
 import com.intellij.psi.util.CachedValuesManager
 import com.intellij.psi.util.PsiTreeUtil.getParentOfType
 import com.mongodb.jbplugin.accessadapter.datagrip.DataGripBasedReadModelProvider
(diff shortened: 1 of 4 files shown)


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
```diff
@@ -139,6 +140,8 @@ abstract class AbstractMongoDbInspectionBridge<Settings, I : Inspection>(
                     queryInspection.run(query, problemsHolder, settings)
                 }
             }
+        } catch (e: ProcessCanceledException) {
```
kmruiz, line 143: IntelliJ linter complains that ProcessCanceledException should be rethrown, not cached,  because it's used to cancel other processes that might be running upwards the stacktrace.

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/editor/CachedQueryService.kt
```diff
@@ -99,12 +99,18 @@ class CachedQueryService(
...
-            CachedValueProvider.Result.create(parsedAst, attachment, connectionStateViewModel)
+            Result.create(parsedAst, attachment, connectionStateViewModel)
         }
 
         attachment.putUserData(queryCacheKey, cachedValue)
-        return decorateWithMetadata(dataSource, attachment.getUserData(queryCacheKey)!!.value)
-    }
+        return@runCatching decorateWithMetadata(dataSource, attachment.getUserData(queryCacheKey)!!.value)
+    }.onFailure { ex ->
+        logger.info(
```
himanshusinghs, line 108:
```suggestion
        logger.warn(
```
Perhaps warn would be better here?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/inspections/AbstractMongoDbInspectionBridge.kt
himanshusinghs: Should we extend the try to also include generation of allQueriesInFile?

himanshusinghs (approved):
Looks solid! I left some non-blocker suggestions.

PR #225, opened 2025-06-05
chore: poc of invalid projections linter INTELLIJ-338

kmruiz:
### Checklist


kmruiz:
The issue is that we don't fetch indexes, we try to run the explain plan of the query. If the query is invalid, the server rejects it. My concern with doing validation in our side is that we might need to keep it up to date with changes on the server and I don't feel that is that beneficial.

himanshusinghs:
Agreed on that part but then the UX is a bit cryptic - such queries I think comes up as dynamic query in the status inlay and then the query generation is also another point where we are generating wrong queries. Don't you think?

PR #226, opened 2025-06-05
chore: ignore sort operations in the type mismatch linter INTELLIJ-333

kmruiz:
### Checklist

### DIFF
--- a/packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/correctness/TypeMismatchInspection.kt
+++ b/packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/correctness/TypeMismatchInspection.kt
@@ -12,6 +12,8 @@ import com.mongodb.jbplugin.mql.CollectionSchema
 import com.mongodb.jbplugin.mql.Node
 import com.mongodb.jbplugin.mql.adt.Either
 import com.mongodb.jbplugin.mql.components.HasFieldReference
+import com.mongodb.jbplugin.mql.components.Name
+import com.mongodb.jbplugin.mql.components.Named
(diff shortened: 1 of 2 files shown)


packages/mongodb-mql-engines/src/main/kotlin/com/mongodb/jbplugin/linting/correctness/TypeMismatchInspection.kt
```diff
@@ -61,6 +63,7 @@ class TypeMismatchInspection<D> : QueryInspection<
                 val collectionSchema = querySchema.value
 
                 val allFieldsAndValuesResult = allNodesWithSchemaFieldReferences<Source>()
+                    .map { node -> node.filter { it.component<Named>()?.name != Name.SORT } }
```
himanshusinghs, line 66: I think it would be better to avoid all the schema and field references that are not in HasFilters component.

kmruiz: I don't think that's advisable: we have other places where we want to do type checking, like on updates (updating a field to a value that is invalid with the current schema).

Right now the only place where we know for sure the type checking is not necessary is for sorting, so I would keep this behaviour we have in the PR.

WDYT?

himanshusinghs: I see. I definitely overlooked that part. Then I would suggest to avoid HasProjections as well and similar ones. Unless that is indirectly covered by `allNodesWithSchemaFieldReferences` looking only at HasValueReference.Constant / HasValueReference.Runtime.

himanshusinghs (approved):

PR #227, opened 2025-06-05
fix(modal): databases were not loading in the run query modal INTELLIJ-340

kmruiz:
Running the events in EDT didn't work because it was waiting for the EDT to be available, and the modal blocks it.


himanshusinghs (approved):
I thought I approved it before :)

PR #242, opened 2025-06-12
chore: do not change the scope to current query when clicking on an inlay INTELLIJ-356

kmruiz:
### Checklist


packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/viewModel/AnalysisScopeViewModel.kt
```diff
@@ -140,28 +139,26 @@ class AnalysisScopeViewModel(
 
         ensureActive()
         log.info(useLogMessage("Analysis in progress...").build())
-        withinReadActionBlocking {
```
kmruiz, line 143: This improves the reliability and the performance of the analysis process. Before this, we were using the inspection tool, which requires access to a read action the whole process. However, the new async API, only needs it when we call apply (print the annotations in the editor). As this is for background files and the editors are closed (and will be open whenever the user navigates to them, when everything will be cached) we can just do all the hard work in a background thread and only the specific bits that require access to the PSI can be done in a `withinReadAction` block.

himanshusinghs (approved):

PR #250, opened 2025-06-13
fix: show no index warnings when a query has no filters INTELLIJ-360

himanshusinghs:
## Description

When a query has no filter, it does not make much sense for us to warn about missing or ineffective index usage. In this PR, we add another check of looking for filters in the query before running the NotUsingIndex and IneffectiveIndexUsage inspections.

Additionally, for queries that do not have any filters, we will now warn about such usage with an insight with title - "Query lacks filters, may fetch the entire collection", and also an inlay called - "No Filters"

### Checklist

### DIFF
--- a/packages/jetbrains-plugin/src/main/resources/messages/InspectionsAndInlaysBundle.properties
+++ b/packages/jetbrains-plugin/src/main/resources/messages/InspectionsAndInlaysBundle.properties
@@ -15,7 +15,7 @@ inlay.indexing.not-run.text=Dynamic Query
 inlay.indexing.not-run.tooltip=Could not analyse the query because it's using unsupported features by the plugin.
 
 inlay.query.not-using-filters.text=No Filters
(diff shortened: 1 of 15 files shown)


packages/jetbrains-plugin/src/main/resources/messages/InspectionsAndInlaysBundle.properties
```diff
@@ -14,6 +14,9 @@ inlay.indexing.ineffective-index-scan.text=Ineffective Index Scan
 inlay.indexing.not-run.text=Dynamic Query
 inlay.indexing.not-run.tooltip=Could not analyse the query because it's using unsupported features by the plugin.
 
+inlay.query.not-using-filters.text=No Filters
+inlay.query.not-using-filters.tooltip=This query lacks filters and will scan the entire collection, which can cause severe performance issues. Either add a filter/match to your query or process the results as a stream/cursor.
```
kmruiz, line 18: My issue with this proposal is that, we will still show the warning if they use a stream or a cursor, right? I'll just suggest to add a filter/match and that's it, so we don't leave the user the feeling that it's broken.

kmruiz (approved):

PR #258, opened 2025-09-16
chore: lazy load icons as composables, skiko might not be loaded at startup now INTELLIJ-364

kmruiz:
### Checklist

### DIFF
--- a/.github/workflows/act-publish-release.yaml
+++ b/.github/workflows/act-publish-release.yaml
@@ -45,7 +45,22 @@ jobs:
 
       - name: Publish Plugin In General Availability
         shell: bash
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          BUILD_SEGMENT_API_KEY: ${{ secrets.BUILD_SEGMENT_API_KEY }}
+          JB_PUBLISH_CHANNEL: "ga"
+          JB_PUBLISH_TOKEN: ${{ secrets.JB_PUBLISH_TOKEN }}
+          JB_CERTIFICATE_CHAIN: ${{ secrets.JB_CERTIFICATE_CHAIN }}
+          JB_PRIVATE_KEY: ${{ secrets.JB_PRIVATE_KEY }}
+          JB_PRIVATE_KEY_PASSWORD: ${{ secrets.JB_PRIVATE_KEY_PASSWORD }}
+        run: |
+          set -e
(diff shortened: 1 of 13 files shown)


.github/workflows/act-publish-release.yaml
```diff
@@ -56,19 +55,34 @@ jobs:
...
           git push origin main
-          
+
           VERSION=$(./gradlew getVersion --quiet --console=plain)
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
           echo "Version: ${VERSION}"
 
       - name: "Notify Release"
         uses: slackapi/slack-github-action@b0fa283ad8fea605de13dc3f449259339835fc52 # 2.1.0
+        if: always()
```
himanshusinghs, line 85: why was this needed? Shouldn't the `always()` be default?

packages/jetbrains-plugin/src/main/kotlin/com/mongodb/jbplugin/ui/components/utilities/ControlledComboBox.kt
```diff
@@ -69,6 +69,7 @@ import kotlinx.coroutines.coroutineScope
 import org.jetbrains.jewel.foundation.Stroke
 import org.jetbrains.jewel.foundation.modifier.border
 import org.jetbrains.jewel.foundation.modifier.onHover
+import org.jetbrains.jewel.foundation.modifier.thenIf
```
himanshusinghs, line 72: is this used somewhere?

packages/jetbrains-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -3,29 +3,23 @@
...
+        <li>IntelliJ version 2025.1 or above</li>
         <li>DataGrip</li>
     </ul>
 
     <b>Features Available<b>
     <ul>
         <li>Write queries faster with autocomplete & type validation for collection and field names in MongoDB Java drivers</li>
-    </ul>
-
-    <b>In Development</b>
```
himanshusinghs, line 23: Should we instead mention in Public Preview? I think I saw this in VSCode marketplace, a label `preview`

himanshusinghs (approved):

packages/jetbrains-plugin/src/main/resources/META-INF/plugin.xml
kmruiz: It's in the header of the description. Do you think we should add it somewhere else?

kmruiz:
I'm going to merge and we can have a follow up PR later :D. This way we can still release and have a version with the fix in the marketplace this week.

